PR #199, opened 2022-06-20
SA-19399/new cancel spinner

kajes:
This implements the new spinner logic from #198  on the `upgrade cancel` command, fixing SA-19399 for this particular command.

The code has been manually tested in windows `cmd.exe` and `Windows PowerShell` and proved to be reliable, not getting stuck or giving any false results from the command.

Other commands will be handled in other PR:s.

Needs #198 

E2E test ID: 696a5bc1-823a-4d9d-a324-b1ed6e6f448c

### DIFF
--- a/cmd/appliance/upgrade/cancel.go
+++ b/cmd/appliance/upgrade/cancel.go
@@ -139,10 +139,13 @@ func upgradeCancelRun(cmd *cobra.Command, args []string, opts *upgradeCancelOpti
 				appliancepkg.UpgradeStatusReady,
 				appliancepkg.UpgradeStatusFailed,
 			}
+			// statusReport is the channel that triggers updates in progress bars
+			statusReport = make(chan string)
 			// wg is the wait group for the progressbars
 			wg           sync.WaitGroup
 			errorChannel = make(chan error, count)
 		)
+		defer close(statusReport)
 		cancelProgressBars := mpb.New(mpb.WithOutput(spinnerOut), mpb.WithWaitGroup(&wg))
 		wg.Add(count)
 		retryCancel := func(ctx context.Context, appliance openapi.Appliance) error {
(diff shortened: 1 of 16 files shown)


cmd/appliance/upgrade/cancel.go
```diff
@@ -139,65 +144,56 @@ func upgradeCancelRun(cmd *cobra.Command, args []string, opts *upgradeCancelOpti
...
 			return backoff.Retry(func() error {
 				return a.UpgradeCancel(ctx, appliance.GetId())
 			}, backoff.NewExponentialBackOff())
 		}
 
-		for _, ap := range appliances {
-			appliance := ap
-			qw.Push(appliance)
-			statusReport := make(chan string)
-			go a.UpgradeStatusWorker.Watch(ctx, cancelProgressBars, appliance, appliancepkg.UpgradeStatusIdle, appliancepkg.UpgradeStatusReady, statusReport)
```
danijeel, line 158: fyi, it would have been enough to set new context here to resolve the issue, everything is superfluous

cmd/appliance/upgrade/cancel.go
```diff
@@ -139,65 +144,56 @@ func upgradeCancelRun(cmd *cobra.Command, args []string, opts *upgradeCancelOpti
 				appliancepkg.UpgradeStatusReady,
 				appliancepkg.UpgradeStatusFailed,
 			}
-			// wg is the wait group for the progressbars
-			wg           sync.WaitGroup
-			errorChannel = make(chan error, count)
 		)
-		cancelProgressBars := mpb.New(mpb.WithOutput(spinnerOut), mpb.WithWaitGroup(&wg))
```
danijeel, line 146: i don't understand why wg is removed, wouldn't it be better to resolve the underlying issue instead? we only loose control.

cmd/appliance/upgrade/cancel.go
```diff
@@ -139,65 +144,56 @@ func upgradeCancelRun(cmd *cobra.Command, args []string, opts *upgradeCancelOpti
...
+		var bars *tui.Progress
+		if !opts.ciMode {
+			bars = tui.NewProgress(ctx, spinnerOut)
+			defer bars.Wait(3 * time.Second)
+		}
+		for _, appliance := range appliances {
+			var s chan<- string
+			if !opts.ciMode {
+				var t *tui.Tracker
+				t, s = bars.AddTracker(appliance.GetName(), "cancelled")
```
danijeel, line 168:
```suggestion
				t, s := bars.AddTracker(appliance.GetName(), "cancelled")
```

kajes: will not work. the `s` chan variable need to passed down, either as nil or as a channel, so we cannot declare `s` as a new variable.

cmd/appliance/upgrade/cancel.go
kajes: I have been testing with wait groups a lot, and it never produces the result we want. Either the spinners get completed too early or not at all.

thomascellerier (approved):

PR #215, opened 2022-06-28
 Add timestamps to important output events

kajes:
This will prepend some of the important event logs with timestamps.


kajes:
E2E: https://got-testrail.agi.appgate.com/index.php?/plans/view/1025351

PR #219, opened 2022-06-30
fix: no upload progress bar in ci-mode

kajes:
This will prevent the progress bar for the file upload progress to execute in ci-mode. It also fixes some styling in the upload bar to unify it's looks with the other spinners.
- changed the upload progress bar to only execute when not using `ci-mode`
- made some styling changes to unify looks of progress bar and spinner trackers
- removed unused function


keathmilligan (approved):
lgtm

Not related to this PR, but I did notice when testing that if you ctrl-c while a tracker is active, the terminal will be left in NECHO state and need to be reset manually. I'll file a separate issue on that.

PR #220, opened 2022-06-30
fix: disambiguity in prepare message

kajes:
This fixes a disambiguity in the prepare message where the use of both "Online" and "Offline" was combined with a "yes" or "no". This makes the language conflict where offline appliances would be presented as "Offline [no]", which could be interpreted as it actually being in an online state.

- all states are either presented as "Online [yes|no]"
- message is also converted to use non-utf8 characters in the windows build
- added more symbols to account for "yes|no" on windows systems


keathmilligan (approved):
lgtm

PR #222, opened 2022-07-01
fix: enable the backup API

kajes:
The `appliance backup api` command was broken in #195 due to forgetting to actually set the boolean to enable the backup API before making the request. This fixes it.

Fixes:
- SA-19614


userappgate (commented):
?

This thing has always worked for me.

PR #225, opened 2022-07-04
activate ci-mode when output is not TTY

kajes:
This sets the `ci-mode` flag if the output is not a TTY. It also moves the TTY detection up in the call stack to make it actually apply earlier.

Fixes:
- SA-19408


kajes:
E2E: https://got-testrail.agi.appgate.com/index.php?/plans/view/1027271

thomascellerier (approved):

PR #226, opened 2022-07-04
5.2 backwards compatability

kajes:
This fixes backwards compatability with api-version 13 (appliance version 5.2). Api-version 13 does not have support for dev-keyring, so preparing an upgrade image on appliance version 5.2 would fail because of this.. This fixes it by:
1. only setting the dev-keyring value for the prepare request if it is true
2. Checking the api-version in the prepare command and setting dev-keyring to false if it v13 or below.


kajes:
E2E: https://got-testrail.agi.appgate.com/index.php?/plans/view/1027353

PR #248, opened 2022-07-20
handling all errs

danijeel:
correctly handle inital version check err and subsequental supressed errors
the inital version check err is not fatal, but should not be supressed.


danijeel:
e2e OK

PR #257, opened 2022-08-17
Fix partial upgrade with filter fails

kajes:
This simplifies the version check when preparing an upgrade and accounts for all appliance versions instead of just the primary controllers version.

Fixes: SA-19776


kajes:
E2E: https://got-testrail.agi.appgate.com/index.php?/plans/view/1045411

danijeel (approved):

PR #264, opened 2022-08-24
removed version missmatch check

danijeel:
this check does not serve any practial use anymore, since https://github.com/appgate/sdpctl/commit/44f913a12a57c5ce5bedaac82badd9c5fe5ba3f5 was introduced.


danijeel:
1048527

kajes:
> this check does not serve any practial use anymore, since https://github.com/appgate/sdpctl/commit/44f913a12a57c5ce5bedaac82badd9c5fe5ba3f5 was introduced.

Not entirely true. There's no guarantee that the re-configuring block from https://github.com/appgate/sdpctl/commit/44f913a12a57c5ce5bedaac82badd9c5fe5ba3f5 will actually execute, since we exit from the function when any upgrade error occurs before it is executed.

We could, however, defer the execution of the re-configuration, so that it will run regardless of any other upgrade errors, but only after we make sure the primary controller upgrade was successful.

danijeel:
e2e OK 1049497

cmd/root.go
```diff
@@ -258,9 +258,23 @@ func rootPersistentPreRunEFunc(f *factory.Factory, cfg *configuration.Config) fu
...
+			// we wont use this check for all commands, and fallback to the config value
+			// so we can reduce number oh http requests to the controller.
+			if configuration.NeedUpdatedAPIVersionConfig(cmd) {
+				minMax, err := auth.GetMinMaxAPIVersion(f)
+				if err == nil && minMax != nil {
+					viper.Set("api_version", minMax.Max)
+					f.Config.Version = int(minMax.Max)
+					if err := viper.WriteConfig(); err != nil {
+						fmt.Fprintf(f.StdErr, "[error] %s\n", err)
+					}
```
kajes, line 274: Not sure if this needs to be written to the config. There are probably cases where the user chooses a specific api-version for some reason. Even if we need the latest api-version, the user may not want a permanent change to the config.

kajes (approved):

PR #265, opened 2022-08-24
clear credentials when configuring

kajes:
Clear credentials when running `sdpctl configure`. This will prevent any old tokens that may present to not be re-used before the user has logged in after re-configuring. It will also force the user to re-enter user credentials before any other command requiring the user to be signed in can be run.

Fixes: SA-19900

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -8,9 +8,11 @@ import (
 	"time"
 
 	"github.com/appgate/sdpctl/pkg/keyring"
+	"github.com/appgate/sdpctl/pkg/util"
 	"github.com/denisbrodbeck/machineid"
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 type Config struct {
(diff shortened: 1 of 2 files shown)


kajes:
E2E: https://got-testrail.agi.appgate.com/index.php?/plans/view/1048643

pkg/configuration/configuration.go
```diff
@@ -168,6 +170,19 @@ func (c *Config) ClearCredentials() error {
...
+		if util.InSlice(k, allKeys) {
+			viper.Set(k, "")
+		}
+	}
+	if err := viper.WriteConfig(); err != nil {
+		// only return error if there is a config file to write to
+		if !errors.As(err, &viper.ConfigFileNotFoundError{}) {
+			return err
+		}
+	}
```
danijeel, line 185: why is this needed? Can you prepend a inline comment why this is needed

kajes: You mean the file not found check? Well, here's what happens if you don't check for it:

danijeel: I just meant add a inline code comment in this function explaining why we need this logic, so our future self does not get confused about it (speaking mainly for myself here)

kajes: Sure, I'll do that

danijeel (approved):

PR #267, opened 2022-08-25
add 2 hour padding on expires_at

danijeel:
To mitigate getting HTTP 401 mid upgrade, we make sure we have atleast 2 hours valid token.


danijeel:
e2e OK 1049173

thomascellerier (approved):

PR #268, opened 2022-08-25
print error details in tracker when prepare/complete fails

kajes:
Tracker now prints the error details when an upgrade prepare/complete fails for some reason, which will help the user know the error before the command exits.

### DIFF
--- a/pkg/appliance/upgrade_status.go
+++ b/pkg/appliance/upgrade_status.go
@@ -69,7 +69,7 @@ func (u *UpgradeStatus) upgradeStatus(ctx context.Context, appliance openapi.App
 			if util.InSlice(s, undesiredStatuses) {
 				if tracker != nil {
 					// send error details for tracker
-					tracker.Update(details)
+					tracker.Update(s + " - " + details)
 				}
 				return backoff.Permanent(fmt.Errorf("Upgrade failed on %s - %s", name, details))
 			}
--- a/pkg/appliance/upgrade_status.go
+++ b/pkg/appliance/upgrade_status.go
@@ -67,6 +67,10 @@ func (u *UpgradeStatus) upgradeStatus(ctx context.Context, appliance openapi.App
 				tracker.Update(s)
 			}
 			if util.InSlice(s, undesiredStatuses) {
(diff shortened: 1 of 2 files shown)


pkg/appliance/upgrade_status.go
```diff
@@ -67,6 +67,10 @@ func (u *UpgradeStatus) upgradeStatus(ctx context.Context, appliance openapi.App
 				tracker.Update(s)
 			}
 			if util.InSlice(s, undesiredStatuses) {
+				if tracker != nil {
+					// send error details for tracker
+					tracker.Update(details)
```
danijeel, line 72: should it be a concat string of  `s + details` maybe? how does this look in practice?

kajes: It looks like this:
```bash
[...]
[2022-08-26T08:09:48+02:00] Uploading upgrade image:
 ✓ appgate-6.0.1-29609-beta.img.zip: upload complete

[2022-08-26T08:11:06+02:00] Preparing image on appliances:
 ⨯ controller-d7e397ee-b8c6-4829-82bf-5ebe22ff1faf-site1: failed - Signature incorrect
 ⨯ gateway-d7e397ee-b8c6-4829-82bf-5ebe22ff1faf-site1: failed - Signature incorrect
2 errors occurred:
	* Upgrade failed on controller-d7e397ee-b8c6-4829-82bf-5ebe22ff1faf-site1 - Signature incorrect
	* Upgrade failed on gateway-d7e397ee-b8c6-4829-82bf-5ebe22ff1faf-site1 - Signature incorrect
```
There's a concat in the tracker code, so the status is always going to be "failed" and then the reason is going to get concatenated like `"failed" + " - " + reason`. Thinking of it now, it probably makes more sense to concatenate it here though, since the status isn't always going to be "failed".

kajes:
E2E: https://got-testrail.agi.appgate.com/index.php?/plans/view/1049457

thomascellerier (approved):
Nice, sometimes the upgrade can continue for quite a while so its good to see this information ASAP.

PR #273, opened 2022-09-01
Fix nil pointer error when no build number is detected

kajes:
This fixes an issue where the `upgrade prepare` would crash with a nil pointer reference error. The error occurs when the image that is being uploaded is the same appliance version as the already present image and build number is missing from the image file name.

Manually tested on Linux and Windows (PS and CMD)

### DIFF
--- a/cmd/appliance/upgrade/prepare.go
+++ b/cmd/appliance/upgrade/prepare.go
@@ -215,7 +215,10 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions) 
 	}
 	if !opts.forcePrepare {
 		var skip []openapi.Appliance
-		appliances, skip = appliancepkg.CheckVersions(ctx, *initialStats, appliances, targetVersion)
+		appliances, skip, err = appliancepkg.CheckVersions(ctx, *initialStats, appliances, targetVersion)
+		if err != nil {
+			return err
+		}
 		if len(appliances) <= 0 {
 			return errors.New("No appliances to prepare for upgrade. All appliances are already greater or equal to the upgrade image")
 		}
--- a/cmd/appliance/upgrade/prepare.go
+++ b/cmd/appliance/upgrade/prepare.go
@@ -157,7 +157,7 @@ func NewPrepareUpgradeCmd(f *factory.Factory) *cobra.Command {
 func checkImageFilename(i string) error {
 	// Check if its a valid filename
 	if rg := regexp.MustCompile(`(.+)?\d\.\d\.\d(.+)?\.img\.zip`); !rg.MatchString(i) {
-		return errors.New("Invalid mimetype on image file. The format is expected to be a .img.zip archive with a version number, such as 5.5.1")
+		return errors.New("Invalid name on image file. The format is expected to be a .img.zip archive with a version number, such as 5.5.1")
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


kajes:
E2E OK: https://got-testrail.agi.appgate.com/index.php?/plans/view/1052057

pkg/appliance/checks_test.go
```diff
@@ -288,6 +288,24 @@ func TestCompareVersionAndBuildNumber(t *testing.T) {
 			v2:   "6.0.0-beta+23456",
 			want: IsGreater,
 		},
+		{
+			desc: "v2 no build number",
```
userappgate, line 292: should this not test for 0 length?

pkg/appliance/checks.go
```diff
@@ -200,7 +200,8 @@ func CompareVersionsAndBuildNumber(x, y *version.Version) int {
 	res := y.Compare(x)
 
 	// if res is 0, we also compare build number
-	if res == IsEqual {
+	// both x and y needs to have a parsable build number for this check to run
+	if res == IsEqual && len(y.Metadata()) > 0 && len(x.Metadata()) > 0 {
 		buildX, _ := version.NewVersion(x.Metadata())
 		buildY, _ := version.NewVersion(y.Metadata())
```
danijeel, line 206: This PR does not solve the issue, you need to check if ptr is nil or handle the error here.

pkg/appliance/checks_test.go
```diff
@@ -347,3 +365,24 @@ func TestHasDiffVersions(t *testing.T) {
...
+	type args struct {
+		x *version.Version
+		y *version.Version
+	}
+	tests := []struct {
+		name string
+		args args
+		want int
+	}{
+		// TODO: Add test cases.
```
danijeel, line 379: ❓

pkg/appliance/checks_test.go
kajes: added some tests for that

pkg/appliance/checks_test.go
kajes: gone now

kajes:
E2E OK: https://got-testrail.agi.appgate.com/index.php?/plans/view/1052131

cmd/appliance/upgrade/prepare.go
```diff
@@ -214,7 +214,10 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions)
 	}
 	if !opts.forcePrepare {
 		var skip []openapi.Appliance
-		appliances, skip = appliancepkg.CheckVersions(ctx, *initialStats, appliances, targetVersion)
+		appliances, skip, err = appliancepkg.CheckVersions(ctx, *initialStats, appliances, targetVersion)
+		if err != nil {
+			return err
```
danijeel, line 219: Is this a fatal error? If so, consider updating the error message; since the error message string from `version.NewVersion` would not be very helpful for anyone other then the developer.

if its not a fatal error, don't `return err`, handle it.

based on this; https://github.com/appgate/sdpctl/blob/4988c37dd77ba66ad5e7116f89505e2ced022a37/pkg/appliance/checks.go#L174-L175 

it does not seems like its treated as severe error.

kajes: I'm in two minds about that. On one hand, failing to parse a version doesn't seem that big of a deal, but on the other hand, we do not encourage downgrading at all, so from that perspective, it's pretty important that we know which version is being uploaded and treat it as fatal if it's a downgrade.

kajes:
E2E: https://got-testrail.agi.appgate.com/index.php?/plans/view/1052547

pkg/appliance/checks.go
```diff
@@ -160,10 +160,15 @@ func ShowAutoscalingWarningMessage(templateAppliance *openapi.Appliance, gateway
 	return tpl.String(), nil
 }
 
+type SkipStruct struct {
```
danijeel, line 163:
```suggestion
type SkipUpgrade struct {
```
name with `Struct` stutter and is superfluous

pkg/appliance/checks.go
```diff
@@ -172,10 +177,26 @@ func CheckVersions(ctx context.Context, stats openapi.StatsAppliancesList, appli
...
+					skip = append(skip, SkipStruct{
+						Appliance: appliance,
+						Reason:    "failed to parse version from stats",
+					})
+					continue
+				}
+				res, err := CompareVersionsAndBuildNumber(statV, v)
+				if err != nil {
+					log.Warn("failed to compare versions")
+					skip = append(skip, SkipStruct{
```
danijeel, line 189: no need to log and return error.  remove log line here and :179-180 .

If you really want to log the value, log once in caller instead of duplicate logs here.

danijeel (changes requested):

pkg/appliance/checks.go
kajes: We do not return the error here though, so not duplicate at all. The error just means we skip this appliance and the only case where we would return these errors is if all the appliances are skipped for some reason. If that does not happen, this is the only place where we actually log the reason, otherwise it's only printed in the summary.

danijeel (approved):

PR #275, opened 2022-09-02
Error 500 response

danijeel:
Handle HTTP 5XX same as 4xx errors, since they may also return the same response body.


danijeel:
e2e OK 1053719

kajes (approved):

PR #278, opened 2022-09-06
Fix tracker error spamming on multi-line failure reasons

kajes:
This fixes an issue where the progress trackers would start spamming the CLI when upgrade prepare fails and the failure reason has line breaks.

Failure reasons are now sanitized before shown in the tracker. Only the first line in the failure details are shown in the tracker and any trailing special characters, such as ':', are removed. The full failure details are printed as errors once the command exits with errors.

This also introduces a dedicated channel and method for sending failure details for the tracker to show.


kajes:
E2E OK: https://got-testrail.agi.appgate.com/index.php?/plans/view/1054103

danijeel (approved):

PR #279, opened 2022-09-06
upgrade complete summary text, respect offline appliances

danijeel:
dont return err if a appliance is offline


danijeel:
e2e 1054139

kajes (approved):

PR #280, opened 2022-09-07
wait for not busy after disabling maintenance mode

kajes:
`upgrade complete` now waits for additional controller to be in a non-busy status before reporting the upgrade as complete. This prevents the command to report an additional controller as upgraded while it still is disabling maintenance mode.

Fixes: SA-2005

### DIFF
--- a/cmd/appliance/upgrade/complete.go
+++ b/cmd/appliance/upgrade/complete.go
@@ -600,16 +600,22 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
 					return err
 				}
 			}
-			if err := a.ApplianceStats.WaitForApplianceState(ctx, controller, appliancepkg.StatReady, t); err != nil {
-				log.WithFields(f).WithError(err).Error("Controller never reached desired state")
-				return err
-			}
 			if cfg.Version >= 15 {
-				_, err := a.DisableMaintenanceMode(ctx, controller.GetId())
-				if err != nil {
+				if err := a.ApplianceStats.WaitForApplianceState(ctx, controller, appliancepkg.StatReady, t); err != nil {
+					return err
+				}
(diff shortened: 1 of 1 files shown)


kajes:
e2e ok https://got-testrail.agi.appgate.com/index.php?/plans/view/1054641

cmd/appliance/upgrade/complete.go
```diff
@@ -600,16 +600,22 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
...
-				log.WithFields(f).WithError(err).Error("Controller never reached desired state")
-				return err
-			}
 			if cfg.Version >= 15 {
-				_, err := a.DisableMaintenanceMode(ctx, controller.GetId())
-				if err != nil {
+				if err := a.ApplianceStats.WaitForApplianceState(ctx, controller, appliancepkg.StatReady, t); err != nil {
+					return err
+				}
+				if _, err := a.DisableMaintenanceMode(ctx, controller.GetId()); err != nil {
```
danijeel, line 607:
```
# https://github.com/appgate/sdpctl/commit/0841a55af60f408a300455e0d9f9bd0331fadd19
git cherry-pick 0841a55af60f408a300455e0d9f9bd0331fadd19
```
and
```golang
changeID, err := a.DisableMaintenanceMode(ctx, controller.GetId())
// ..
// ..
change, err := ac.RetryUntilCompleted(ctx, changeID, controller.GetId())
```

thomascellerier (approved):
I suppose there is some timeout?

cmd/appliance/upgrade/complete.go
kajes: Added :+1:

kajes:
> I suppose there is some timeout?

yes, it uses the same timeout as the one used for the upgrade itself.

PR #281, opened 2022-09-08
maintenance toggle

danijeel:
New command: Toggle maintenance mode on a single controller.

image


danijeel:
I'd like to get some input on the docs and wording here...

thomascellerier:
Regarding wording:

I would use "disable maintenance mode" rather than "set maintenance mode to false".
Same for "enable maintenance mode" rather than "set maintenance mode to true".

On the description: "An appliance in maintenance mode won't allow to perform POST, PUT, PATCH or DELETE methods.
This is a superuser function and should only be used if you know what you are doing."

Right now maintenance mode is only applicable to Controllers though the API currently let's you set it on any appliance (something to fix?).
Maintenance mode behaves differently in 5.5 and 6.0. In 6.0 clients are still able to connect.

So maybe something like:

A Controller in maintenance mode will not accept any API calls besides disabling maintenance mode. Starting in version 6.0, clients will still function as usual while a Controller is in  maintenance mode.
This is a superuser function and should only be used if you know what you are doing.

The version dependent information might be logged depending on API version?

pkg/docs/maintenance.go
```diff
@@ -0,0 +1,109 @@
+package docs
+
+var (
+	MaintenanceModeLongDescription = `
+    This function is a advance function, and should be used with care,
```
kajes, line 5:
```suggestion
    For advanced users only.
```

kajes (changes requested):
Is the toggle command really necessary if we have enable and disable commands? I find it a bit superfluous and confusing.

thomascellerier:
Its fine by me to remove the toggle sub-command, you would generally either want to explicitly enable or disable and having separate sub commands for that makes mistakes harder.

thomascellerier (approved):
Tried it out looks good.

kajes (approved):

Issue #241, opened 2022-07-07
sdpctl continues with upgrade process in case of cancelled upgrade


kajes:
sdpctl does not check volume number after an upgrade has been completed on an appliance before continuing.

Internal ref: SA-19359

PR #283, opened 2022-09-12
Check volume number when completing upgrade

### LINKED ISSUE #240: testing: tests should account for different responses on same endpoint
The sdpctl testing suite should be able to give different responses on multiple requests to the same endpoint.

Currently, the API mocking does not allow for registering multiple responses for the same endpoint, which makes it difficult to test for changing appliance states.
For example, before an upgrade, an appliance has a volume number. Upon a successful upgrade, this volume number will change, indicating that the appliance has switched partition to the new version. In short, if the volume number hasn't changed after `upgrade complete` has finished on the appliance, that would indicate a failed or cancelled upgrade.

The volume number is fetched from the appliance stats API and checking the volume number means that the stats API endpoint need to called at least twice for each appliance that is upgraded. This means testing this scenario is impossible due to the test suite not allowing multiple responses to register for the same endpoint.

kajes:
Add checks for volume number after upgrade completes. If volume number is the same as before the upgrade complete is executed, the upgrade will fail and return the error.

Also fixes breaking tests by adding a new `MutatingResponse` function that allows the test suite to re-use an endpoint, but get different results. The response mutation is done through a callback function provided to the test function.

Fixes #240 and #241

### DIFF
--- a/pkg/httpmock/http.go
+++ b/pkg/httpmock/http.go
@@ -121,6 +121,30 @@ func setup() (*openapi.APIClient, *openapi.Configuration, *http.ServeMux, *httpt
 	return c, clientCfg, mux, server, port, server.Close
 }
 
+func MutatingResponse(filename string, callback func(count int, i []byte) ([]byte, error)) http.HandlerFunc {
+	count := 0
+	return func(w http.ResponseWriter, r *http.Request) {
+		f, err := os.Open(filename)
+		if err != nil {
+			panic(fmt.Sprintf("Internal testing error: could not open %q", filename))
+		}
+		defer f.Close()
+		reader := bufio.NewReader(f)
+		content, err := io.ReadAll(reader)
+		if err != nil {
+			panic(fmt.Sprintf("Internal testing error; could not read %q", filename))
+		}
+		mutated, err := callback(count, content)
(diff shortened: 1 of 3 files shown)


kajes:
e2e ok https://got-testrail.agi.appgate.com/index.php?/plans/view/1056849

pkg/httpmock/http.go
```diff
@@ -121,6 +121,30 @@ func setup() (*openapi.APIClient, *openapi.Configuration, *http.ServeMux, *httpt
 	return c, clientCfg, mux, server, port, server.Close
 }
 
+func MutatingResponse(filename string, callback func(count int, i []byte) ([]byte, error)) http.HandlerFunc {
+	count := 0
```
danijeel, line 125: You probably should use a mutex here, I think this will hinder parallel tests using this func.

kajes: Probably yes, I'll add it in.

thomascellerier: How is the test failing? deadlock?

kajes: yep, deadlock.

danijeel (approved):

PR #291, opened 2022-09-15
sdpctl profiles, Mange more then 1 appgate sdp collective

danijeel:
Add a new, more user focused approach to handle more then one appgate sdp collective configuration.
Internal issue tracker: SA-19912

Next up:
- ~~Documentation and command descriptions will come in a follow-up PR to reduce size of PR.~~
- Seperate log file per collective (Internal issue SA-19914)

https://user-images.githubusercontent.com/1218385/190962867-bbc8ea4c-9b88-461f-855a-4b4f3342ce50.mp4


danijeel:
e2e OK 1058329

danijeel:
going to refactor and rename from `collective` to `profile`

thomascellerier:
I'm reviewing just the UX here and not the actual code.

First of all I think this looks really solid and well designed.

Here is some general feedback gathered on this:
- I would say `sdpctl profile` is a more appropriate name since you can gave several profiles per collective.
- It would be nice to have a -p/--profile top level flag to sdpctl to select profiles explicitly for a single command. For example `sdpctl -p foo appliance list`.
- In the same maner it would be good to have an enviornment variable for the same thing, e.g. `SDPCTL_PROFILE=foo sdpctl appliance list`.

These last 2 points are especially useful when scripting/automating things via sdpctl.

cmd/profile/add.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/appgate/sdpctl/pkg/docs"
+	"github.com/appgate/sdpctl/pkg/filesystem"
+	"github.com/appgate/sdpctl/pkg/profiles"
+	"github.com/appgate/sdpctl/pkg/util"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"golang.org/x/sync/errgroup"
+)
+
+var validProfileName = regexp.MustCompile(`^[A-Za-z0-9_.]+$`)
```
kajes, line 21: I think dashes (-) should be valid as well.

cmd/profile/add.go
```diff
@@ -0,0 +1,216 @@
...
+
+// NewAddCmd return a new profile add command
+func NewAddCmd(opts *commandOpts) *cobra.Command {
+	return &cobra.Command{
+		Use:   "add [<name>]",
+		Short: docs.ProfileAddDoc.Short,
+		Long:  docs.ProfileAddDoc.Long,
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
```
kajes, line 30: We could use a prompt if no profile name is given as argument

cmd/profile/add.go
```diff
@@ -0,0 +1,216 @@
...
+		return err
+	}
+	fmt.Fprintf(opts.Out, "Created profile %s, run 'sdpctl profile list' to see all available profiles\n", name)
+	fmt.Fprintf(opts.Out, "run 'sdpctl profile set %s' to select the new profile profile\n", name)
+	return nil
+}
+
+// updatePemPath if we have a existing pem file in the default config, update the pem_filepath
+// to respect the new directory tree
+func updatePemPath(targetDir string, files []string) error {
```
kajes, line 121: Is this needed? `pem_filepath` is converted to an absolute path in configure.go#60

pkg/docs/profile.go
```diff
@@ -0,0 +1,39 @@
...
+		Long:  `Mange local configuration files for sdpctl towards more then one appgate sdp profile.`,
+	}
+	ProfileAddDoc = CommandDoc{
+		Short: "Add another appgate sdp profile configuration",
+		Long:  `Add creates a new appgate sdp profile configuration directory`,
+	}
+	ProfileDeleteDoc = CommandDoc{
+		Short: "Remove an existing profile profile",
+		Long:  `Remove an existing profile profile from your local configuration settings`,
+	}
```
kajes, line 15: These should have some examples as well.

cmd/profile/set.go
```diff
@@ -0,0 +1,75 @@
...
+	if !profiles.FileExists() {
+		fmt.Fprintln(opts.Out, "no profiles added")
+		fmt.Fprintln(opts.Out, "run 'sdpctl profile add' first")
+		return nil
+	}
+	p, err := profiles.Read()
+	if err != nil {
+		return err
+	}
+	length := len(p.List)
```
kajes, line 36: If I'm already using a default profile and create a new profile, once I set the new profile, I can't go back to my default profile again. It should be in the list of selectable profiles, but it's not added in the `profiles.json` file.

kajes (changes requested):
A few comments, most notably the default profile is not selectable once you add more profiles. Would also like to have the ability to rename a profile as well.

cmd/profile/set.go
danijeel: I can't reproduce, or I don't understand. Could you elaborate?

danijeel:
> A few comments, most notably the default profile is not selectable once you add more profiles. 

can't reproduce this, Can you elaborate? 


> Would also like to have the ability to rename a profile as well.

Won't fix that in this PR. We can do it later if the need arise.

cmd/profile/set.go
kajes: Ah, there seem to some wierdness here. Here's how to reproduce if you start with only having the `$SDPCTL_CONFIG_DIR/config.json` from before:
```bash
# initialise by trying to add a new profile called 'default'
> sdpctl profile add default
1 error occurred:
	* profile already exists with the name default

# list profiles and see that it's empty
> sdpctl profile ls
no current profile is set, run 'sdpctl profile set'

Available profiles
Name    Config directory
----    ----------------

# add a second profile
> sdpctl profile add secondary
Created profile secondary, run 'sdpctl profile list' to see all available profiles
run 'sdpctl profile set secondary' to select the new profile profile

# run ls again, still no default profile
> sdpctl profile ls
no current profile is set, run 'sdpctl profile set'

Available profiles
Name         Config directory
----         ----------------
secondary    /home/larskajes/.config/sdpctl/profiles/secondary
```

This seems very edge case, but I bet we're going to have some user run into this if not fixed :P.

cmd/profile/set.go
```diff
@@ -0,0 +1,75 @@
...
+			PageSize: length,
+			Message:  "select profile:",
+			Options:  list,
+		}
+		if err := prompt.SurveyAskOne(qs, &index, survey.WithValidator(survey.Required)); err != nil {
+			return err
+		}
+	}
+	p.Current = &p.List[index].Directory
+	fmt.Fprintf(opts.Out, "%s (%s) is selected as current sdp profile profile\n", p.List[index].Name, p.List[index].Directory)
```
kajes, line 65: 'profile' seem to double up here
```
> sdpctl profile set secondary
secondary (/home/larskajes/.config/sdpctl/profiles/secondary) is selected as current sdp profile profile
```

kajes (changes requested):

kajes (approved):
Looks good now :+1:

thomascellerier (approved):
Very nice :+1:

PR #293, opened 2022-09-19
Add command for managing service users

kajes:
This implements a new sdpctl command for managing service users on Appgate SDP Collective. It has support for creating, reading (list and get), updating and deleting service users. The command also has support for disabling and enabling service users.

Creating and updating service users can be done either by prompt, flags or by reading a pre-defined JSON file. The UX here could probably use some work though.

**Basic usage:**
```bash
# creating a user
>sdpctl service-users create
? Name for service user: <service-user-name>
? Passphrase for service user: <passphrase>
? Confirm your passphrase: <passphrase>
New service user created:
{
  [...]
  "id": "<uuid>",
  "name": "<service-user-name>",
  [...]
}

# create user non-interactively
> echo <passphrase> | sdpctl service-users create --name <service-user-name>
New service user created:
[...]

# create from file
> sdpctl service-users create --from-file=<path-to-json-file>
New service user created:
[...]

# listing users
> sdpctl service-users list

# disable user
>sdpctl service-users update <uuid> disable

# removing a user
> sdpctl service-users remove <uuid>
```

**Demo:**
[asciicast](https://asciinema.org/a/zv4SSCS3NGXD1Re7uJpQ8qNbh)

Fixes: SA-19919


kajes:
e2e ok: https://got-testrail.agi.appgate.com/index.php?/plans/view/1061662

danijeel (changes requested):
You need to add a version check on all subcommands here if serviceusers is available

pkg/docs/serviceusers.go
```diff
@@ -0,0 +1,118 @@
...
+				Command:     "sdpctl service-users list",
+			},
+			{
+				Description: "output in JSON format",
+				Command:     "sdpctl service-users list --json",
+			},
+		},
+	}
+	ServiceUsersGet = &CommandDoc{
+		Short: "get detailed infomation about a service user",
```
danijeel, line 24:
```suggestion
		Short: "get detailed information about a service user",
```

thomascellerier:
I will test it out.

thomascellerier:
Some feedback, we should not support passing passphrases via command line arguments.
It should be environment variables and even better via stdin/file descriptior.

Typo here:
```
  # create service user with flag input
  > sdpctl service-users create --name=<service-user-name> --passphrase=<service-user-passphrase
```

kajes:
> Some feedback, we should not support passing passphrases via command line arguments. It should be environment variables and even better via stdin/file descriptior.

The flag can be easily removed, but there''s support for passing in password via sdtin already:
```bash
echo <passphrase> | sdpctl service-users create --name <name>
```

thomascellerier:
Ah nice I missed that. Yes I think we shouldn't have password flags in general.

cmd/serviceusers/create.go
```diff
@@ -0,0 +1,171 @@
...
+						if err != nil {
+							return err
+						}
+					}
+
+				}
+
+				var errs *multierror.Error
+				if len(username) <= 0 {
+					errs = multierror.Append(errors.New("name is required"), errs)
```
danijeel, line 131: wrong order on Append arguments, see https://github.com/hashicorp/go-multierror#usage

cmd/serviceusers/create.go
```diff
@@ -0,0 +1,171 @@
...
+				}
+				if len(password) <= 0 {
+					errs = multierror.Append(errors.New("passphrase is required"), errs)
```
danijeel, line 134: wrong order on Append arguments, see https://github.com/hashicorp/go-multierror#usage

cmd/serviceusers/create.go
```diff
@@ -0,0 +1,171 @@
...
+				}
+				if errs != nil {
+					errs = multierror.Append(errors.New("failed to create user: missing data"), errs)
```
danijeel, line 137: wrong order on Append arguments, see https://github.com/hashicorp/go-multierror#usage

cmd/serviceusers/create.go
```diff
@@ -0,0 +1,171 @@
...
+
+			if len(users) <= 0 {
+				return fmt.Errorf("failed to create user(s): no user data provided")
+			}
+
+			var errs *multierror.Error
+			for _, u := range users {
+				created, err := api.Create(ctx, u)
+				if err != nil {
+					errs = multierror.Append(err, errs)
```
danijeel, line 155: wrong order on Append arguments, see https://github.com/hashicorp/go-multierror#usage

cmd/serviceusers/delete.go
```diff
@@ -0,0 +1,98 @@
...
+		Use:     "delete [id...]",
+		Short:   docs.ServiceUsersDelete.Short,
+		Long:    docs.ServiceUsersDelete.Long,
+		Example: docs.ServiceUsersDelete.ExampleString(),
+		Aliases: []string{"remove", "rm", "del"},
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			var errs *multierror.Error
+			for _, arg := range args {
+				if !util.IsUUID(arg) {
+					errs = multierror.Append(fmt.Errorf("%s: %s", InvalidUUIDError, args[0]), errs)
```
danijeel, line 32: wrong order on Append arguments, see https://github.com/hashicorp/go-multierror#usage

cmd/serviceusers/list.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	cmd := &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   docs.ServiceUsersList.Short,
+		Long:    docs.ServiceUsersList.Long,
+		Example: docs.ServiceUsersList.ExampleString(),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			var err error
+			if opts.JSON, err = cmd.Flags().GetBool("json"); err != nil {
```
danijeel, line 27: why is this block needed?

cmd/serviceusers/create.go
```diff
@@ -0,0 +1,171 @@
...
+		API:    f.ServiceUsers,
+		In:     f.Stdin,
+		Out:    f.IOOutWriter,
+	}
+	cmd := &cobra.Command{
+		Use:     "create",
+		Short:   docs.ServiceUsersCreate.Short,
+		Long:    docs.ServiceUsersCreate.Long,
+		Example: docs.ServiceUsersCreate.ExampleString(),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
danijeel, line 36: I think you can reduce indentation level in this function significant if you refactor and return earlier

cmd/serviceusers/update.go
```diff
@@ -0,0 +1,169 @@
...
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 {
+				return fmt.Errorf("not enough arguments")
+			}
+			if !util.IsUUID(args[0]) {
+				return fmt.Errorf(InvalidUUIDError)
+			}
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
```
danijeel, line 38: I think you can reduce indentation level significantly here if you refactor and return earlier(/and or extract funcs)

cmd/serviceusers/update_test.go
```diff
@@ -0,0 +1,197 @@
...
+				if !tC.wantOut.MatchString(parsedOut) {
+					t.Fatalf("\n%s\n != \n%s", tC.wantOut.String(), parsedOut)
+				}
+			}
+		})
+	}
+}
+
+func defaultUpdateResponseHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method == http.MethodGet {
```
danijeel, line 164: we can reduce the size signfiicantly here
for example

```golang
func defaultUpdateResponseHandler(w http.ResponseWriter, r *http.Request) {
	var filename string
	if r.Method == http.MethodGet {
		filename := "../../pkg/serviceusers/fixtures/service-user-get.json"
	}
	if r.Method == http.MethodPut {
		filename := "../../pkg/serviceusers/fixtures/service-user-update.json"
	}
	f, err := os.Open(filename)
	if err != nil {
		panic(fmt.Sprintf("Internal testing error: could not open %q", filename))
	}
	defer f.Close()
	w.Header().Set("Content-Type", "application/json")
	w.WriteHeader(http.StatusOK)
	reader := bufio.NewReader(f)
	content, err := io.ReadAll(reader)
	if err != nil {
		panic(fmt.Sprintf("Internal testing error: could not read %q", filename))
	}
	fmt.Fprint(w, string(content))
}

```

kajes: Makes a whole lot of sense

kajes:
e2e: https://got-testrail.agi.appgate.com/index.php?/plans/view/1063849

cmd/serviceusers/create.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	cmd := &cobra.Command{
+		Use:     "create",
+		Short:   docs.ServiceUsersCreate.Short,
+		Long:    docs.ServiceUsersCreate.Long,
+		Example: docs.ServiceUsersCreate.ExampleString(),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if opts.Config.Version <= 16 {
+				return fmt.Errorf("The service user interface is only available from API version 17 or higher. Currently using API version %d", opts.Config.Version)
+			}
```
danijeel, line 39: I think it would be enough to append a `PersistentPreRunE` func in `service-users` (serviceusers.go) instead of doing this 5 times in every subcommand.

cmd/serviceusers/create.go
```diff
@@ -0,0 +1,176 @@
...
+			return fmt.Errorf("file not found: %s", path)
+		}
+		file, err := os.ReadFile(path)
+		if err != nil {
+			return err
+		}
+		// Try unmarshal json array first
+		dtoArray := ServiceUserArrayDTO{}
+		if err := json.Unmarshal(file, &dtoArray); err != nil {
+			logrus.WithError(err).Warn("failed to unmarshal json to array")
```
danijeel, line 84: this seems un-nessesary to log this, given this flow for example:

```bash
sdpctl service-users create --no-interactive --from-file users.json
WARN[0000] failed to unmarshal json to array             error="json: cannot unmarshal object into Go value of type serviceusers.ServiceUserArrayDTO"
New service user created:
{
  "created": "2022-09-28T06:25:43.256229Z",
  "disabled": false,
  "failedLoginAttempts": 0,
  "id": "e4818be0-eb89-4180-809e-01f1aafefb27",
  "labels": {},
  "name": "daniel",
  "notes": "",
  "tags": [],
  "updated": "2022-09-28T06:25:43.256229Z"
}
~                                                                                                                                                                                                           8:25
> echo $?                                       
0
~                                                                                                                                                                                                           8:25
> jq . users.json
{
  "name": "daniel",
  "password": "abc123"
}

```

cmd/serviceusers/update.go
```diff
@@ -0,0 +1,176 @@
...
+		Aliases: []string{"edit", "set"},
+		Args: func(cmd *cobra.Command, args []string) error {
+			if opts.Config.Version <= 16 {
+				return fmt.Errorf("The service user interface is only available from API version 17 or higher. Currently using API version %d", opts.Config.Version)
+			}
+			if len(args) < 1 {
+				return fmt.Errorf("not enough arguments")
+			}
+			if !util.IsUUID(args[0]) {
+				return fmt.Errorf(InvalidUUIDError)
```
danijeel, line 37:
```suggestion
				return fmt.Errorf("%s: %s", InvalidUUIDError, args[0])
```

pkg/docs/serviceusers.go
```diff
@@ -0,0 +1,118 @@
...
+		Short: "delete one or more service user(s)",
+		Long:  "",
+		Examples: []ExampleDoc{
+			{
+				Description: "delete a service user with the id of <id>",
+				Command:     "sdpctl service-users delete <id>",
+			},
+			{
+				Description: "delete multiple service users by providing multiple id:s",
+				Command:     "sdpctl service-users delete <id1> id2>",
```
danijeel, line 110:
```suggestion
				Command:     "sdpctl service-users delete <id1> <id2>",
```

cmd/serviceusers/update.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	toUpdate.SetTags(dto.Tags)
+	toUpdate.SetLabels(dto.Labels)
+
+	updated, err := api.Update(ctx, *toUpdate)
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(opts.Out, "Updated service user %s:\n", updated.GetName())
```
danijeel, line 174: this can cause problems, for example
```
 sdpctl service-users update --no-interactive e4818be0-eb89-4180-809e-01f1aafefb27 '{"name": "NewName"}' --json | jq .
parse error: Invalid numeric literal at line 1, column 8
```
if i pass the flag `--json` i expect only json output, but I get mixed output.

cmd/serviceusers/create.go
kajes: That would overwrite the `PersistentPreRunE` from root though, which is kind of important. That's why it's repeated.

cmd/serviceusers/create.go
kajes: removed it

cmd/serviceusers/create.go
kajes: So instead, I implemented a version restriction using command annotation instead and do the actual checking in the root `PersistentPreRunE`. I think it may be useful to have for cases like these.

pkg/configuration/configuration.go
```diff
@@ -91,6 +92,21 @@ func IsAuthCheckEnabled(cmd *cobra.Command) bool {
...
+	c := cmd
+	for c != nil {
+		if c.Annotations != nil {
+			if s, ok := c.Annotations["MinAPIVersion"]; ok {
+				if minVersion, err := strconv.ParseInt(s, 10, 64); err == nil && int64(currentVersion) < minVersion {
+					return fmt.Errorf("Minimum API version %d is required to use the '%s' command. Current API version is %d", minVersion, c.Name(), currentVersion)
+				}
+			}
+		}
+		c = c.Parent()
```
danijeel, line 105: is this some debug leftover?

kajes: No, this is what makes it recurse back up the command tree, so that we can set the annotation on the `service-users` command and it will apply to all sub commands as well.

pkg/configuration/configuration.go
```diff
@@ -91,6 +92,21 @@ func IsAuthCheckEnabled(cmd *cobra.Command) bool {
 	return true
 }
 
+func CheckMinAPIVersionRestriction(cmd *cobra.Command, currentVersion int) error {
+	c := cmd
+	for c != nil {
+		if c.Annotations != nil {
+			if s, ok := c.Annotations["MinAPIVersion"]; ok {
+				if minVersion, err := strconv.ParseInt(s, 10, 64); err == nil && int64(currentVersion) < minVersion {
```
danijeel, line 100:
```suggestion
				if minVersion, err := strconv.Atoi(s); err == nil && currentVersion < minVersion {
```
the api version is a predictable integer value, we could use Atoi here for brevity, to avoid typecasting.

kajes: makes sense

danijeel (approved):

PR #295, opened 2022-09-20
FIX: clear auth token when running signin

kajes:
Clear any old token that is currently stored when running `configure signin`. This prevents sdpctl from trying to use any token that is being stored, even though it has expired. In effect, the `configure signin` command will now always renew the authentication.

Fixes: SA-20111

e2e ok: https://got-testrail.agi.appgate.com/index.php?/plans/view/1060763

### DIFF
--- a/pkg/keyring/keyring_windows.go
+++ b/pkg/keyring/keyring_windows.go
@@ -125,8 +125,6 @@ func DeleteBearer(prefix string) error {
 	if _, ok := os.LookupEnv("SDPCTL_BEARER"); ok {
 		os.Unsetenv("SDPCTL_BEARER")
 	}
-	viper.Set("bearer", "")
-	viper.Set("expires_at", "")
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


kajes:
@mandopaloooza Would you mind testing this out on your Mac as well?

pkg/keyring/keyring_windows.go
```diff
@@ -116,6 +116,18 @@ func SetBearer(prefix, secret string) error {
 	return saveEncryptedFile(bearer, prefix, secret)
 }
 
+func DeleteBearer(prefix string) error {
```
danijeel, line 119: This wont work, the bearer token is not stored in the keyring, since the size is to big, on windows its stored as a encrypted dpapi file on disk, I think what you want todo is this
https://github.com/appgate/sdpctl/blob/31162b8dfd2fa8ecdd1b8de47e0ae919a3f4a73d/pkg/keyring/keyring_windows.go#L34-L42

kajes: ops! I thought that was a Mac specific problem. Anyway, made a quick PR here #296

PR #296, opened 2022-09-21
delete bearer file in windows

kajes:
fix for #295 windows issue

- remove bearer token in windows properly
- refactor commonly used functions

### DIFF
--- a/pkg/keyring/keyring.go
+++ b/pkg/keyring/keyring.go
@@ -21,13 +21,16 @@ const (
 // ClearCredentials removes any existing items in the keychain,
 // it will ignore if not found errors
 func ClearCredentials(prefix string) error {
-	for _, k := range []string{username, password, bearer} {
+	for _, k := range []string{username, password, bearer, refreshToken} {
 		if err := deleteSecret(format(prefix, k)); err != nil {
 			if !errors.Is(err, zkeyring.ErrNotFound) {
 				return err
 			}
 
 		}
+	}
+	if _, ok := os.LookupEnv("SDPCTL_BEARER"); ok {
+		os.Unsetenv("SDPCTL_BEARER")
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


danijeel (changes requested):
we should refactor here and combine this logic with a private method https://github.com/appgate/sdpctl/blob/495c5499c79cf6153961b4078f247b5a6edb63c4/pkg/keyring/keyring_windows.go#L26 instead of copy paste

pkg/keyring/keyring_darwin.go
```diff
@@ -141,7 +141,7 @@ func SetBearer(prefix, secret string) error {
 }
 
 func DeleteBearer(prefix string) error {
-	if err := deleteSecret(prefix, bearer); err != nil {
+	if err := deleteSecretKey(prefix, bearer); err != nil {
```
danijeel, line 144: I'm not sure this should be a fatal error here, since these keys does not always exists

pkg/keyring/keyring.go
```diff
@@ -21,14 +21,17 @@ const (
 // ClearCredentials removes any existing items in the keychain,
 // it will ignore if not found errors
 func ClearCredentials(prefix string) error {
-	for _, k := range []string{username, password, bearer} {
+	for _, k := range []string{username, password, bearer, refreshToken} {
```
danijeel, line 24: im not sure we want to remove refreshToken since this make refreshToken obsolete and defeats the purpose

pkg/keyring/keyring.go
```diff
@@ -86,6 +89,15 @@ func SetRefreshToken(prefix, secret string) error {
 	return setSecret(format(prefix, refreshToken), secret)
 }
 
+func DeleteRefreshToken(prefix string) error {
```
danijeel, line 92: when is this method used?

pkg/keyring/keyring_windows.go
```diff
@@ -24,21 +24,15 @@ import (
...
 	}
-	p, err := filepath.Abs(fmt.Sprintf("%s/%s", profiles.GetStorageDirectory(), format(prefix, bearer)))
-	if err != nil {
-		return err
-	}
-	if ok, err := util.FileExists(p); err == nil && ok {
-		if err := os.Remove(p); err != nil {
-			return err
-		}
+	if _, ok := os.LookupEnv("SDPCTL_BEARER"); ok {
```
danijeel, line 34: i was expecting you to call the `DeleteBearer() `method from inside `ClearCredentials()`, how come it was removed?

kajes: Sorry, can't remember really what my thought was, but I added it now.

danijeel (approved):

PR #298, opened 2022-09-23
FIX: ParseInt error when comparing upload version and prepared version

kajes:
This fixes a bug where `sdpctl appliance upgrade prepare` would throw a `ParseInt` error if the already prepared version does not have a build number to compare with.


kajes:
e2e ok: https://got-testrail.agi.appgate.com/index.php?/plans/view/1062110

danijeel (approved):

PR #301, opened 2022-09-27
Always log version when running command

kajes:
Added a log entry with useful debug information that is logged on every command run.

Fixes SA-20133

### DIFF
--- a/cmd/appliance/upgrade/complete.go
+++ b/cmd/appliance/upgrade/complete.go
@@ -114,6 +114,7 @@ func NewUpgradeCompleteCmd(f *factory.Factory) *cobra.Command {
 }
 
 func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeCompleteOptions) error {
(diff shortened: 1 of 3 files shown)


danijeel:
Based on the feedback I've seen, I think it would be useful to display the current version of sdpctl  to stdout when running `sdpctl appliance upgrade prepare` and `sdpctl appliance upgrade complete`

kajes:
> Based on the feedback I've seen, I think it would be useful to display the current version of sdpctl to **stdout** when running `sdpctl appliance upgrade prepare` and `sdpctl appliance upgrade complete`

In that case, we might as well print that for every command. I'm guessing that if we only print it for those two, there will definitely be requests for printing it for the others as well.

danijeel:
>In that case, we might as well print that for every command. I'm guessing that if we only print it for those two, there will definitely be requests for printing it for the others as well.

no this will be a very bad change, this will hinder the usage of sdpctl in conjunction with other projects with pipes.

kajes:
Should be better now.

danijeel (approved):

PR #302, opened 2022-09-28
Check backup file result before trying to download

kajes:
If the generation of the backup file fails for some reason, sdpctl would not know about it and try to download the file even if it fails on the server.

This implements a check for the result of the backup file generation, implemented in API version 17, to prevent the backup command from trying to download failed backup files.

Fixes SA-19909

### DIFF
--- a/pkg/appliance/backup.go
+++ b/pkg/appliance/backup.go
@@ -223,7 +223,7 @@ func PerformBackup(cmd *cobra.Command, args []string, opts *BackupOpts) (map[str
 			log.WithFields(log.Fields{
 				"appliance": applianceID,
 				"backup_id": backupID,
-				"status":    status,
+				"status":    status.GetStatus(),
 			}).Info("backup status")
 
 			if status.GetStatus() != backup.Done {
--- a/pkg/appliance/backup.go
+++ b/pkg/appliance/backup.go
@@ -246,12 +246,15 @@ func PerformBackup(cmd *cobra.Command, args []string, opts *BackupOpts) (map[str
 		if err := retryStatus(ctx, b.applianceID, b.backupID); err != nil {
 			return b, err
 		}
-		result, output, err := backupAPI.Result(ctx, b.applianceID, b.backupID)
-		if err != nil {
-			return b, err
-		}
-		if result != backup.Success {
-			return b, errors.New(output)
+		if opts.Config.Version >= 17 {
+			// Check for backup result and output implemented in API version 17
+			result, output, err := backupAPI.Result(ctx, b.applianceID, b.backupID)
+			if err != nil {
+				return b, err
+			}
+			if result != backup.Success {
+				return b, errors.New(output)
+			}
 		}
 		log.WithFields(f).Infof("starting download for backup id %s", b.backupID)
 		file, err := backupAPI.Download(ctx, b.applianceID, b.backupID)
--- a/pkg/appliance/backup.go
+++ b/pkg/appliance/backup.go
@@ -246,6 +246,13 @@ func PerformBackup(cmd *cobra.Command, args []string, opts *BackupOpts) (map[str
 		if err := retryStatus(ctx, b.applianceID, b.backupID); err != nil {
 			return b, err
 		}
+		result, output, err := backupAPI.Result(ctx, b.applianceID, b.backupID)
+		if err != nil {
+			return b, err
+		}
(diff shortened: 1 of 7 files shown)


kajes:
e2e ok: https://got-testrail.agi.appgate.com/index.php?/plans/view/1064915

pkg/appliance/backup.go
```diff
@@ -246,6 +246,16 @@ func PerformBackup(cmd *cobra.Command, args []string, opts *BackupOpts) (map[str
...
 			return b, err
 		}
+		if opts.Config.Version >= 17 {
+			// Check for backup result and output implemented in API version 17
+			result, output, err := backupAPI.Result(ctx, b.applianceID, b.backupID)
+			if err != nil {
+				return b, err
+			}
+			if result != backup.Success {
+				return b, errors.New(output)
```
danijeel, line 256: I'd rather see that refactor backup.Status instead of .Result() method, since **output** can be empty, this error message does not really make since.

how about
```golang
	retryStatus := func(ctx context.Context, applianceID, backupID string) error {
		bo := backoff.NewExponentialBackOff()
		bo.MaxElapsedTime = 0
			status, err := backupAPI.Status(ctx, applianceID, backupID)
			if err != nil {
				return err
			}
			log.WithFields(log.Fields{
				"appliance":     applianceID,
				"backup_id":     backupID,
				"backup_status": status.GetStatus(),
			}).Infof("backup status %s", status.GetOutput()) // output as string, not field, since it can be empty

			if status.GetStatus() != backup.Done {
				return fmt.Errorf("Backup not done for appliance %s, got %s %s %s", applianceID, status.GetStatus(), status.GetOutput(), status.GetResult())
			}
			return nil
		}, bo)
	}
```


and backup.status
```diff
diff --git a/pkg/appliance/backup/backup.go b/pkg/appliance/backup/backup.go
index 5d3457a..87b8133 100644
--- a/pkg/appliance/backup/backup.go
+++ b/pkg/appliance/backup/backup.go
@@ -59,12 +59,12 @@ const (
 	Failure    string = "failure"
 )
 

-func (b *Backup) Status(ctx context.Context, applianceID, backupID string) (string, error) {
+func (b *Backup) Status(ctx context.Context, applianceID, backupID string) (*openapi.AppliancesIdBackupBackupIdStatusGet200Response, error) {
 	status, response, err := b.APIClient.ApplianceBackupApi.AppliancesIdBackupBackupIdStatusGet(ctx, applianceID, backupID).Authorization(b.Token).Execute()
 	if err != nil {
-		return "", api.HTTPErrorResponse(response, err)
+		return nil, api.HTTPErrorResponse(response, err)
 	}
-	return status.GetStatus(), nil
+	return status, nil
 }
 
 func (b *Backup) Result(ctx context.Context, applianceID, backupID string) (string, string, error) {
```

pkg/appliance/backup.go
```diff
@@ -246,6 +246,16 @@ func PerformBackup(cmd *cobra.Command, args []string, opts *BackupOpts) (map[str
 		if err := retryStatus(ctx, b.applianceID, b.backupID); err != nil {
 			return b, err
 		}
+		if opts.Config.Version >= 17 {
```
danijeel, line 249: we don't have to worry about version checking here since the new fields are omitted if empty

pkg/appliance/backup.go
kajes: I can refactor the `backup.Status`, but not this way because it would not work. We need to first check if the status is done _and then_ what the result is. Only checking the status is not enough since it can be done, but still have failed. That's what the result field is for.

pkg/appliance/backup.go
kajes: It is also omitted in API version 17<= _if_ the status is not done, which is stated in the spec. Either way, we definitely need some check, either an API version check _or_ a nil check, and since there are more cases where the result field is omitted even on 17<=, a version check is better IMHO.

danijeel:
this is a nice improvement, however it does not resolve the internal ticket SA-19909, (it asks for support for both the new fields, result and  the **message** field

kajes:
> this is a nice improvement, however it does not resolve the internal ticket SA-19909, (it asks for support for both the new fields, result and the **message** field

That field does not seem to be included in the client code yet, so that would be a separate PR once the client has that field as well.

thomascellerier (approved):

PR #303, opened 2022-09-28
skip write error on api version update

danijeel:
This error would normally only be triggered if running sdpctl with a headless environment (such as only configured with environment variable) the program works fine without updating the config file.

we will only show error if we do not get 'Config File "config" Not Found %dir'


danijeel:
e2e 1064517

PR #306, opened 2022-09-30
add quick start guide as github page

gokhan-berberoglu:


kajes (approved):
If there are no other changes to be made, I think we can merge this.

PR #312, opened 2022-10-11
Validate hostname uniqueness

danijeel:


danijeel:
e2e OK 1070427

pkg/network/network.go
```diff
@@ -0,0 +1,50 @@
...
+	// if ipv6 is not supported on the host.
+	ipv4, _ := resolver.LookupIP(ctx, "ip4", addr)
+	ipv6, _ := resolver.LookupIP(ctx, "ip6", addr)
+
+	if len(ipv4) > 1 || len(ipv6) > 1 {
+		return &ResolverError{
+			ip:       append(ipv4, ipv6...),
+			hostname: addr,
+		}
+	}
```
thomascellerier, line 48: You could also fail if there are 0 ips.

thomascellerier (approved):
Note that we also have a check for that in the appliance upgrade code itself starting with 6.0.3.

PR #314, opened 2022-10-12
Appliance seed file

danijeel:
New command, `sdpctl appliance export-seed `
Get the seed file for a inactive appliance in json or iso format.

this PR includes some minor refactor to utilize common functionality from other commands, review commit-by-commit.

internal ticket: SA-20122


cmd/appliance/seed.go
```diff
@@ -0,0 +1,276 @@
...
+			return nil
+		},
+		RunE: func(c *cobra.Command, args []string) error {
+			return seedRun(c, args, &opts)
+		},
+	}
+	cmd.Flags().BoolVar(&opts.provideCloudSSHKey, "provide-cloud-ssh-key", false, "Tells appliance to use the key generated by AWS or Azure")
+	cmd.Flags().BoolVar(&opts.allowCustomization, "allow-customization", false, "Whether the Appliance should allow customizations or not")
+	cmd.Flags().BoolVar(&opts.iso, "iso-format", false, "Export as ISO format")
+	cmd.Flags().BoolVar(&opts.json, "json", false, "output as json to stdout, only applicable in json seed type")
```
kajes, line 171: Maybe we should provide an `--output|o` flag if they the user wants a file created instead?

cmd/appliance/seed.go
```diff
@@ -0,0 +1,276 @@
...
+	cmd.Flags().StringVarP(&opts.sshKey, "ssh-key", "", "", "filepath to public ssh key to be allowed")
+	cmd.Flags().IntVarP(&opts.validityDays, "validity-days", "", 1, "How many days the seed should be valid for")
```
kajes, line 173: This could be a time duration flag as well, which would provide a lot more flexibility. Of course that would be useless if the seed validity can only be n days.

cmd/appliance/seed.go
```diff
@@ -0,0 +1,276 @@
...
+		if err != nil {
+			return err
+		}
+		sshConfig.SetSshKey(pub)
+	} else if opts.provideCloudSSHKey {
+		sshConfig.SetProvideCloudSSHKey(true)
+	} else if len(opts.password) < 1 {
+		hasStdin := false
+		stat, err := os.Stdin.Stat()
+		if err == nil && (stat.Mode()&os.ModeCharDevice) == 0 {
```
kajes, line 217: We do this check in other places as well. We should extract this as a function

cmd/appliance/seed.go
```diff
@@ -0,0 +1,276 @@
...
+		ext  string
+	)
+	if opts.iso {
+		seed, response, err := a.APIClient.AppliancesApi.AppliancesIdExportIsoPost(ctx, appliance.GetId()).SSHConfig(*sshConfig).Authorization(token).Execute()
+		if err != nil {
+			return api.HTTPErrorResponse(response, err)
+		}
+		data = []byte(seed.GetIso())
+		ext = "iso"
+	} else {
```
kajes, line 240: Since there are flags for both json and iso, I think we should explicitly check for the json flag value instead. This logic implies that json is the default format, which would make the json flag obsolete.

cmd/appliance/seed.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	path, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	outputfile := filepath.Join(path, fmt.Sprintf("%s_seed.%s", appliance.GetId(), ext))
+	if err := os.WriteFile(outputfile, data, 0644); err != nil {
+		return err
+	}
```
kajes, line 264: So the output is written to a file anyway? In that case, the description for the json flag is wrong.

cmd/appliance/seed.go
```diff
@@ -0,0 +1,276 @@
...
+	} else {
+		sshConfig.SetPassword(opts.password)
+	}
+
+	var (
+		data []byte
+		ext  string
+	)
+	if opts.iso {
+		seed, response, err := a.APIClient.AppliancesApi.AppliancesIdExportIsoPost(ctx, appliance.GetId()).SSHConfig(*sshConfig).Authorization(token).Execute()
```
kajes, line 234: I think we should be put this client logic in the `pkg/api.go` file instead. Will be easier to maintain if know where to find these call functions to begin with.

pkg/docs/appliance.go
```diff
@@ -285,4 +285,48 @@ NOTE: Although the '--include' and '--exclude' flags are provided as options her
...
+    Appliance customizations are disabled by default. These can be enabled later using a cz-config command if required.
+
+SSH Authentication Method:
+    --provide-cloud-ssh-key
+    Use SSH key provided by the cloud instance
+    With Cloud based appliances, you should specify a key when launching the instance.
+
+    --ssh-key
+    Use SSH public key
+    Begins with ssh-rsa, ssh-dss, ssh-ed25519, ecdsa-sha2-nistp256, ecdsa-sha2-nistp384 or ecdsa-sha2-nistp521.
```
kajes, line 308: Help for the flags are already in the flag description provided by the command function. This is duplicate information.

kajes (changes requested):
Running only `appliance export-seed` will only print `please provide options to select from`. I would expect it to print the help message.

kajes (approved):

PR #318, opened 2022-10-18
Add circuit breaker when polling for backup status in case of network errors

kajes:
This will check for network errors when polling the status of an initiated backup by using a timeout context. The timeout for each request is set to 5 seconds. If a request times out 5 consecutive times, it will be considered a permanent error and result in the a command error.

The network error counter will be reset if the status polling request is successful before reaching the 5 counter threshold.

Fixes: SA-20059

### DIFF
--- a/pkg/appliance/backup.go
+++ b/pkg/appliance/backup.go
@@ -213,9 +213,16 @@ func PerformBackup(cmd *cobra.Command, args []string, opts *BackupOpts) (map[str
 	retryStatus := func(ctx context.Context, applianceID, backupID string, tracker *tui.Tracker) error {
 		bo := backoff.NewExponentialBackOff()
 		bo.MaxElapsedTime = 0
+		networkErrors := 0
 		return backoff.Retry(func() error {
 			status, err := backupAPI.Status(ctx, applianceID, backupID)
 			if err != nil {
+				if errors.Is(err, context.DeadlineExceeded) {
+					networkErrors++
+					if networkErrors >= 5 {
+						return backoff.Permanent(err)
+					}
+				}
 				return err
 			}
 			log.WithFields(log.Fields{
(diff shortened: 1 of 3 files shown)


pkg/appliance/backup.go
```diff
@@ -213,11 +213,19 @@ func PerformBackup(cmd *cobra.Command, args []string, opts *BackupOpts) (map[str
...
 		bo := backoff.NewExponentialBackOff()
 		bo.MaxElapsedTime = 0
+		networkErrors := 0
 		return backoff.Retry(func() error {
 			status, err := backupAPI.Status(ctx, applianceID, backupID)
 			if err != nil {
+				if errors.Is(err, context.DeadlineExceeded) {
+					networkErrors++
+					if networkErrors >= 5 {
+						return backoff.Permanent(err)
```
danijeel, line 223: I think it would be suitable to have a human friendly error message here if this happends, since context deadline error is not very user friendly.

kajes: Wouldn't this snippet in root.go do just that though? https://github.com/appgate/sdpctl/blob/c1b18d676c320cfa179acf44810270f612167aff/cmd/root.go#L211-L213

danijeel: no, since this err is handled within the command, https://github.com/appgate/sdpctl/blob/3d4972f1a876fe36beb6b1cb9fabec8d522f3dad/pkg/appliance/backup.go#L293-L295

kajes: Fixed in 4f5d038f3fc5a36961098b9a19f97267bd95c15c

thomascellerier (approved):
This looks good but needs some follow-up work.
We need a way to handle an aborted backup.
We should suggest either resuming or deleting the backup.

This means we also need to expose a command to delete a backup id and also the capability to resume an existing backup. Maybe do that automatically if backup_status is "processing" when starting the backup.

PR #319, opened 2022-10-18
Fix possible race condition when preparing an upgrade image

kajes:
This fixes a possible race condition where sdpctl would start polling for the upgrade status before the appliance starts the process, causing progress bars to fail with no error message.

The idle state is no longer set as an unwanted status while preparing until later in the upgrade process. On API version 15 and above, the idle status will be unwanted once the change request has been completed. On API versions below 15, the idle state will be set as unwanted after download of the upgrade image has completed.

Fixes: SA-20337


thomascellerier (approved):
Looks good, I will try it out.

thomascellerier:
Yes I can confirm this fixes the race condition.

PR #328, opened 2022-11-02
Add --actual-hostname flag for prepare

kajes:
This adds the `--actual-hostname` flag for the prepare command as well, making it validate the hostname uniqueness properly and giving the user an opportunity avoid the error

There's a fair bit of refactoring here too, since I discovered that we had duplicate code logic for this that ran at different times in different places. The important part is really in the `cmd/prepare.go`.


cmd/appliance/upgrade/prepare.go
```diff
@@ -208,9 +207,12 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions)
 	if err != nil {
 		return err
 	}
+	if len(opts.actualHostname) > 0 {
+		host = opts.actualHostname
+	}
```
thomascellerier, line 212: This doesn't override the host actually used to establish the connection Right?
This should just override the host when running `FindPrimaryController` and for the `controller://` file download URL.

kajes: No, it's only used for finding the primary controller and then never used again.

pkg/network/network.go
```diff
@@ -6,6 +6,10 @@ import (
 	"net"
 	"sort"
 	"strings"
+
+	"github.com/appgate/sdp-api-client-go/api/v17/openapi"
```
danijeel, line 10: I think this refactor is a mistake, these imports only invites further problems with circular imports, I'd suggest move it back to appliance pkg.

cmd/appliance/upgrade/complete.go
```diff
@@ -102,21 +94,14 @@ func NewUpgradeCompleteCmd(f *factory.Factory) *cobra.Command {
 			return nil
 		},
 		RunE: func(c *cobra.Command, args []string) error {
-			h, err := opts.Config.GetHost()
-			if err != nil {
-				return fmt.Errorf("could not determine hostname for %s", err)
-			}
-			if err := network.ValidateHostnameUniqueness(h); err != nil {
-				return err
-			}
```
danijeel, line 111: I don't understand why this was removed, and the commit message/pr description does not describe it. how come?

cmd/appliance/upgrade/prepare.go
```diff
@@ -157,6 +155,7 @@ func NewPrepareUpgradeCmd(f *factory.Factory) *cobra.Command {
 	flags.BoolVar(&opts.DevKeyring, "dev-keyring", false, "Use the development keyring to verify the upgrade image")
 	flags.Int("throttle", 5, "Upgrade is done in batches using a throttle value. You can control the throttle using this flag.")
 	flags.BoolVar(&opts.hostOnController, "host-on-controller", false, "Use primary controller as image host when uploading from remote source.")
+	flags.StringVar(&opts.actualHostname, "actual-hostname", "", "If the actual hostname is different from that which you are connecting to the appliance admin API, this flag can be used for setting the actual hostname.")
```
danijeel, line 158: why not mimic the same description as the python script?

cmd/appliance/upgrade/prepare.go
```diff
@@ -140,13 +145,6 @@ func NewPrepareUpgradeCmd(f *factory.Factory) *cobra.Command {
 			return errs
 		},
 		RunE: func(c *cobra.Command, args []string) error {
-			h, err := opts.Config.GetHost()
-			if err != nil {
-				return fmt.Errorf("could not determine hostname for %s", err)
-			}
-			if err := network.ValidateHostnameUniqueness(h); err != nil {
-				return err
-			}
```
danijeel, line 149: same as previous comment, I don't understand how removing this is related to actual hostname changes. This error message is show if the config value for controller API resolves to more then 1 ip, it should not be related to this problem

cmd/appliance/upgrade/complete.go
kajes: Because it's moved. It's run when identifying the primary controller instead

cmd/appliance/upgrade/prepare.go
kajes: It's mimicing the flag used in complete already

cmd/appliance/upgrade/prepare.go
kajes: Same answer as above as well

pkg/network/network.go
kajes: which ones? be more specific about it, because the duplicate `ValidateHostnameUniqueness` function was put in a new `network` package when it was created in the first place.

cmd/appliance/upgrade/prepare.go
thomascellerier: OK we just need to make sure both approaches come up with the same answer.

kajes: the other one depends on the former, so yes

cmd/appliance/upgrade/prepare.go
kajes: oh wait, now I see what you mean, but why are we validating the config host? No-one has asked for that as far as I know.

kajes:
So I reverted some of the refactoring and also the validation of the config hostname.

danijeel:
I think you've been abit too creative with git here, the functions `TestStatsIsOnline`, `GetRealHostname`,` ValidateHostname` looks like they have just been moved around within their respective file, so it makes it hard to follow what has actually changed here. Is there any way we can make the diff output smaller? or is everything new?

kajes:
> I think you've been abit too creative with git here, the functions `TestStatsIsOnline`, `GetRealHostname`,` ValidateHostname` looks like they have just been moved around within their respective file, so it makes it hard to follow what has actually changed here. Is there any way we can make the diff output smaller? or is everything new?

`TestStatsIsOnline` and `GetRealHostname` are unchanged, just moved around in file. In `ValidateHostname`, I only extracted the part that checks if the IP:s are unique, so that it uses the function in `network` instead. The `ValidateHostnameUniqueness` function has changed a bit to account for errors. It's basically a merge between the two duplicated functions that existed.

pkg/network/network.go
```diff
@@ -29,22 +31,34 @@ The hostname resolves to the following ips:
...
-	ipv4, _ := resolver.LookupIP(ctx, "ip4", addr)
-	ipv6, _ := resolver.LookupIP(ctx, "ip6", addr)
-
-	if len(ipv4) > 1 || len(ipv6) > 1 {
+	resolver := net.DefaultResolver
+	ipv4, err := resolver.LookupIP(ctx, "ip4", addr)
+	if err != nil {
+		errCount++
+		err = fmt.Errorf("ipv4: %w", err)
+		errs = multierror.Append(err, errs)
```
danijeel, line 42:
```suggestion
		errs = multierror.Append(errs, fmt.Errorf("ipv4: %w", err))
```
looks like you've mixed up the order of Append,
See https://github.com/hashicorp/go-multierror#usage for example

pkg/network/network.go
```diff
@@ -29,22 +31,34 @@ The hostname resolves to the following ips:
...
+	}
+	ipv6, err := resolver.LookupIP(ctx, "ip6", addr)
+	if err != nil {
+		errCount++
+		err = fmt.Errorf("ipv6: %w", err)
+		errs = multierror.Append(err, errs)
```
danijeel, line 48:
```suggestion
		errs = multierror.Append(errs, fmt.Errorf("ipv6: %w", err))
```
looks like you've mixed up the order of Append,
See https://github.com/hashicorp/go-multierror#usage for example

danijeel (approved):

PR #329, opened 2022-11-03
Remove timeout while uploading

kajes:
If there's a bad connection while uploading a file to a controller, the command would timeout before the file is uploaded.

This fix removes the timeout while uploading entirerly. The timeout on the `upgrade prepare` command will only take effect once the file has been uploaded.

Fixes: SA-20469


thomascellerier (approved):
We could still have a timeout on the individual write calls though.

One thing for "read timeout" is that Controller does some checksum calculation at the end of the upgrade file upload which can make the last read take a while, but we should still be able to have some sort of timeout there. @natan-abolafya Is that still the case?

natan-abolafya:
yes

PR #330, opened 2022-11-04
Auth, read from memory

danijeel:
reset current config values, not viper values.

Converted config values, bearer and expires_at to ptr to avoid future confusion in the subject.

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -225,6 +225,8 @@ func (c *Config) ClearBearer() error {
 	if err := keyring.DeleteBearer(h); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/configuration/configuration.go
```diff
@@ -225,8 +229,9 @@ func (c *Config) ClearBearer() error {
 	if err := keyring.DeleteBearer(h); err != nil {
 		return err
 	}
-	viper.Set("bearer", "")
-	viper.Set("expires_at", "")
```
danijeel, line 229: this was the issue, since it tries to reset config value in file, but seeing how we use it, we want the values that is already read into the struct to be null

kajes (approved):

PR #331, opened 2022-11-07
Cmd switch partition

danijeel:
Internal ticket: SA-20513

### DIFF
--- a/pkg/docs/appliance.go
+++ b/pkg/docs/appliance.go
@@ -323,4 +323,17 @@ for More information, see: https://sdphelp.appgate.com/adminguide/new-appliance.
 			},
 		},
 	}
+
+	ApplianceUpgradeSwitchPartitionDoc = CommandDoc{
+		Short: "Switch partition",
+		Long: `Reboot the appliance and switch the secondary partition to become the primary.
+
+`,
+		Examples: []ExampleDoc{
+			{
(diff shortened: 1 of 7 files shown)


pkg/docs/appliance.go
```diff
@@ -285,4 +285,17 @@ NOTE: Although the '--include' and '--exclude' flags are provided as options her
 			},
 		},
 	}
+
+	ApplianceUpgradeSwitchPartitionDoc = CommandDoc{
+		Short: "Switch partition",
+		Long: `Reboot the appliance and switch the secondary partition to become the primary.
+
```
danijeel, line 292: I'd like some input here, maybe add a big warning description?

Example usage right now
```
> sdpctl appliance upgrade switch-partition 
? select appliance: gateway2-2da97007-d5cc-4db2-b33d-a1e27c79abd6-site1 - Default Site - []
? Are you really sure you want to switch-partition on gateway2-2da97007-d5cc-4db2-b33d-a1e27c79abd6-site1?

Do you want to continue? No
1 error occurred:
	* Cancelled by user

```

kajes: You can add some description as to why you would use this command and why you'd need to be careful about it. Something like `This is useful if...`

kajes: I've updated the docs with a suggestion. The warning message needs to be filled in though.

PR #332, opened 2022-11-07
new command: sdpctl appliance logs

danijeel:
new command: `sdpctl appliance logs`

Download all logs from a specific appliance

Internal ticket: SA-20519

### DIFF
--- a/cmd/appliance/logs.go
+++ b/cmd/appliance/logs.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/appgate/sdpctl/pkg/api"
 	"github.com/appgate/sdpctl/pkg/cmdappliance"
+	"github.com/appgate/sdpctl/pkg/docs"
 	"github.com/appgate/sdpctl/pkg/factory"
 	"github.com/appgate/sdpctl/pkg/terminal"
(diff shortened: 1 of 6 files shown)


thomascellerier (approved):
Did some manual testing, very nice addition :+1:
Not only for this command but in general it would be nice to have an alternative json format output mode to be able to use that in scripts.
In this case the json would contain the absolute path to the downloaded log bundle.

cmd/appliance/logs.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	p := mpb.New(mpb.WithWidth(64), mpb.WithOutput(opts.SpinnerOut()))
+
+	bar := p.New(0,
+		mpb.SpinnerStyle(tui.SpinnerStyle...),
+		mpb.BarFillerOnComplete(tui.Check),
+		mpb.BarWidth(1),
+		mpb.AppendDecorators(
+			decor.Name(name+" ", decor.WCSyncWidthR),
```
kajes, line 117: The `decor.WCSyncWidthR` option here will not do anything since it only applies when you use multiple bars and you can only download from one appliance at a time.

PR #336, opened 2022-11-07
Get controller hostname from config

kajes:
Use the controllers configured hostname when using `--host-on-controller` flag.


thomascellerier (approved):
This covers also the upload case right?

PR #339, opened 2022-11-08
Better UX when filtering

kajes:
The filtering feature will now catch errors that occur due to invalid regex query string and return.

In case all appliances are filtered before showing the upgrade prepare summary, the command will now exit, printing a general error that there are no appliances to prepare for upgrade, as well as listing why each of the appliances was skipped.

The filter will also be applied after excluding all offline appliances, meaning offline appliances will never accidentally be included when using the filtering flags.

Fixes SA-20523

### DIFF
--- a/cmd/appliance/upgrade/prepare.go
+++ b/cmd/appliance/upgrade/prepare.go
@@ -235,7 +235,10 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions) 
 			Appliance: a,
 		})
 	}
-	appliances, filtered := appliancepkg.FilterAppliances(online, filter)
+	appliances, filtered, err := appliancepkg.FilterAppliances(online, filter)
+	if err != nil {
+		return err
+	}
 	for _, f := range filtered {
 		skipAppliances = append(skipAppliances, appliancepkg.SkipUpgrade{
 			Appliance: f,
--- a/cmd/appliance/upgrade/prepare.go
+++ b/cmd/appliance/upgrade/prepare.go
@@ -251,10 +251,10 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions) 
 			var errs *multierr.Error
 			if len(skipAppliances) > 0 {
 				for _, skip := range skipAppliances {
-					errs = multierr.Append(fmt.Errorf("%s skipped: %s", skip.Appliance.GetName(), skip.Reason), errs)
+					errs = multierr.Append(errs, fmt.Errorf("%s skipped: %s", skip.Appliance.GetName(), skip.Reason))
 				}
 			}
-			errs = multierr.Append(errors.New("No appliances to prepare for upgrade. The filter query may be invalid. See the log for more details."), errs)
+			errs = multierr.Append(errs, errors.New("No appliances to prepare for upgrade. The filter query may be invalid. See the log for more details."))
 			return errs
 		}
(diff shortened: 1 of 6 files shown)


thomascellerier (approved):
Nice.

Do you also validate that the passed key is valid? E.g. if I pass naem= instead of name=.

kajes:
> Nice.
> 
> Do you also validate that the passed key is valid? E.g. if I pass naem= instead of name=.

Yes, that would result in a warning message in the log. Of course, you could also account for some level of common typos and accept them. For instance, it's pretty reasonable to think that `naem` would actually be `name`.

thomascellerier:
But does it actually fail the command? A syntax error should just print the error and exit IMO (on top of any logging).

kajes:
> But does it actually fail the command? A syntax error should just print the error and exit IMO (on top of any logging).

Writing `naem` instead of `name` is not a syntax error though, just a typo. Either way, I disagree about making actual syntax errors fatal, since you can combine different kinds keywords as well as the `--include`/`--exclude` flags. You may have a syntax error in one keyword query, which would result in it not matching anything, but not in another, which would still work. On top of that, there's the summary, which lists both the appliances that should be included and the ones that are excluded, and it is up to user to confirm that their queries are actually working the way they should.

Note though that if you only use one keyword query with a syntax error, you would still end up with an error anyway, since it wouldn't match any appliances, eg. nothing to prepare.

cmd/appliance/upgrade/prepare.go
```diff
@@ -247,7 +254,7 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions)
 					errs = multierr.Append(fmt.Errorf("%s skipped: %s", skip.Appliance.GetName(), skip.Reason), errs)
 				}
 			}
-			errs = multierr.Append(errors.New("No appliances to prepare for upgrade."), errs)
+			errs = multierr.Append(errors.New("No appliances to prepare for upgrade. The filter query may be invalid. See the log for more details."), errs)
```
danijeel, line 257: wrong order of arguments, see https://github.com/hashicorp/go-multierror#usage

pkg/appliance/functions_test.go
```diff
@@ -604,7 +604,7 @@ func TestFilterAndExclude(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := FilterAppliances(tt.args.appliances, tt.args.filter); !reflect.DeepEqual(got, tt.want) {
```
danijeel, line 607: i was expect an assertion of the second return value here

cmd/appliance/upgrade/prepare.go
```diff
@@ -247,7 +254,7 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions)
 					errs = multierr.Append(fmt.Errorf("%s skipped: %s", skip.Appliance.GetName(), skip.Reason), errs)
```
danijeel, line 254: same here, wrong order

thomascellerier:
I think any syntax error should result in a failing command since the command will not be applied to the appliances you think they will be applied too. Potentially resulting in a bad/destructive action.
That's also consistent with the behavior of the old upgrade script (though it used glob syntax instead of regexp).

kajes:
@thomascellerier Ok, now it returns on any syntax errors as well

PR #340, opened 2022-11-10
more reliable version identification on prepare

kajes:
Read the local image metadata file to identify which appliance version is being prepared. This is a more reliable way of identifying the version, so will result in less errors due to problems identifying it through file naming.

As a result, most filenames are valid and version parsing errors are now treated as fatal in most cases during prepare.

Fixes SA-20417

### DIFF
--- a/cmd/appliance/upgrade/prepare.go
+++ b/cmd/appliance/upgrade/prepare.go
@@ -1,7 +1,6 @@
 package upgrade
 
 import (
-	"archive/zip"
 	"bytes"
 	"context"
 	"errors"
@@ -56,6 +55,7 @@ type prepareUpgradeOptions struct {
 	forcePrepare     bool
 	ciMode           bool
 	actualHostname   string
+	targetVersion    *version.Version
 }
 
 // NewPrepareUpgradeCmd return a new prepare upgrade command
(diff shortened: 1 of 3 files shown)


pkg/appliance/helpers.go
```diff
@@ -1,24 +1,17 @@
...
-	HelpManualURL = "https://sdphelp.appgate.com/adminguide/v5.5"
-
-	BackupInstructions = `
-Please perform a backup or snapshot of %s before continuing!
-Use appgate-backup to perform a backup of the Controller.
-For more documentation on the backup process, go to:
-    %s/backup-script.html
-`
-)
-
```
kajes, line 21: This was dead code

cmd/appliance/upgrade/prepare.go
```diff
@@ -170,8 +182,8 @@ func NewPrepareUpgradeCmd(f *factory.Factory) *cobra.Command {
 
 func checkImageFilename(i string) error {
 	// Check if its a valid filename
-	if rg := regexp.MustCompile(`(.+)?\d\.\d\.\d(.+)?\.img\.zip`); !rg.MatchString(i) {
-		return errors.New("Invalid name on image file. The format is expected to be a .img.zip archive with a version number, such as 5.5.1")
+	if rg := regexp.MustCompile(`\.img\.zip`); !rg.MatchString(i) {
+		return errors.New("Invalid name on image file. The format is expected to be a .img.zip archive.")
```
kajes, line 186: file naming isn't as strict anymore. The only criteria is that the file ends with `.img.zip`

danijeel (approved):

PR #342, opened 2022-11-10
Upgrade status error messages 

danijeel:
This PR improves error message overall, mainly during upgrade complete.

- append retry on requests that we've seen fail, for various reasons.
- Get spinner upgrade/status error message from context value
- error messages during barchupgrade now include appliance name in the error message
- appliance/change now return earlier if we get "failure" in response body
- mitigate logging nested mutlierr string to logfile

fixes internal issue: SA-20430

### DIFF
--- a/cmd/appliance/upgrade/complete.go
+++ b/cmd/appliance/upgrade/complete.go
@@ -612,7 +612,7 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
 				}
 				logEntry.Info("Install the downloaded upgrade image to the other partition")
 				if !SwitchPartition {
-					if err := a.UpgradeStatusWorker.WaitForUpgradeStatus(context.WithValue(ctx, appliancepkg.UpgradeStatusGetErrorMessage, "applying upgrade"), i, []string{appliancepkg.UpgradeStatusSuccess}, []string{appliancepkg.UpgradeStatusFailed}, t); err != nil {
+					if err := a.UpgradeStatusWorker.WaitForUpgradeStatus(ctx, i, []string{appliancepkg.UpgradeStatusSuccess}, []string{appliancepkg.UpgradeStatusFailed}, t); err != nil {
 						return fmt.Errorf("%s %w", i.GetName(), err)
 					}
 					status, err := a.UpgradeStatusRetry(ctx, i.GetId())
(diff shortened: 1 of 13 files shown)


pkg/appliance/upgrade_status.go
```diff
@@ -27,33 +27,37 @@ var defaultExponentialBackOff = &backoff.ExponentialBackOff{
...
-					tracker.Update("installing")
+				msg := "no response, appliance offline"
+				if !errors.Is(err, context.DeadlineExceeded) {
+					msg = err.Error()
 				}
+				if v, ok := ctx.Value(UpgradeStatusGetErrorMessage).(string); ok {
+					msg = v
+				}
+
+				tracker.Update(msg)
```
kajes, line 60: I don't think using the context to pass messages is a good idea in this case. For starters, the "no response, appliance offline" message will never be shown to the user because it is immediately replaced by the error passed in context on line 57, so L52-L55 is effectively dead code.
But more importantly, we lose a lot of fine grained control over messages here. The rebooting logic that is removed here was there because it was once asked for, and it gives the user a lot more information on what's going on even when the API is unreachable. This change just makes it less informative and reverts a previous feature request, but I don't really understand why this is a necessary change? What does this fix?

cmd/appliance/upgrade/complete.go
```diff
@@ -632,10 +649,13 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
...
 			}
-			if err := a.UpgradeComplete(ctx, controller.GetId(), true); err != nil {
+			err := backoff.Retry(func() error {
+				return a.UpgradeComplete(ctx, controller.GetId(), true)
+			}, backoff.WithContext(backoff.NewExponentialBackOff(), ctx))
+			if err != nil {
 				return err
 			}
-			if err := a.UpgradeStatusWorker.WaitForUpgradeStatus(ctx, controller, []string{appliancepkg.UpgradeStatusIdle}, []string{appliancepkg.UpgradeStatusFailed}, t); err != nil {
+			if err := a.UpgradeStatusWorker.WaitForUpgradeStatus(context.WithValue(ctx, appliancepkg.UpgradeStatusGetErrorMessage, "applying upgrade"), controller, []string{appliancepkg.UpgradeStatusIdle}, []string{appliancepkg.UpgradeStatusFailed}, t); err != nil {
```
kajes, line 658:
```suggestion
			if err := a.UpgradeStatusWorker.WaitForUpgradeStatus(context.WithValue(ctx, appliancepkg.UpgradeStatusGetErrorMessage, "switching partition"), controller, []string{appliancepkg.UpgradeStatusIdle}, []string{appliancepkg.UpgradeStatusFailed}, t); err != nil {
```
This is not different from what we do in the ´batchUpgrade´ function, so the message should be the same.

pkg/appliance/upgrade_status.go
danijeel: > For starters, the "no response, appliance offline" message will never be shown to the user because it is immediately

No this is not true.
 See image for example.
img-2022-11-15-110914
In this case, the appliance got terminated mid operation.

 Upgrade status is viewed both in prepare and complete. it all depends on ctx and what type of error we get.

> But more importantly, we lose a lot of fine grained control over messages here.

I would argue we get even more control with this implementation then the prior. Now we can acctually append correct error message, depending on when the function is being used.


> What does this fix?

It no longer report false error messages, if a appliance is offline/internal error, it would report as "rebooting", but that would not always be true.

kajes: > I would argue we get even more control with this implementation then the prior. Now we can acctually append correct error message, depending on when the function is being used.

So how does this differentiate between a 504 response and no response at all? How do you determine if a reboot has happened or not and present different information depending on that. Like I mentioned, that was a feature that was implemented on request. We can still use the values in context, but that doesn't really explain why the reboot logic was removed. They are not mutually excluding each other.

> It no longer report false error messages, if a appliance is offline/internal error, it would report as "rebooting", but that would not always be true.

That could have been solved in other ways. It should not show "rebooting" on prepare, but on complete when it is actually rebooting. That's a simple boolean switch.

kajes (approved):

PR #347, opened 2022-11-16
Re-work documentation

kajes:
Use auto generated markdown documentation instead of the manually written ones. The markdown documents are placed under the `/docs` folder in the repository and is purely based on documentation provided in code. The markdown files are based on the same texts as the `--help` section for the commands, so they are identical. Each command has its own dedicated file as well.

This avoids the issue of having multiple places where we need to update documentation when needed, meaning less disparate information and less of a headache for developers.


gokhan-berberoglu:
Updated langauge with the following convention:

- Do not use "Appgate SDP" within the text, it is redundant.
   - for example instead of "the Appgate SDP Collective" use "the Collective"
- Start with uppercase for Appgate SDP specific terms like Collective, Controller, LogServer etc
   - appliance is a generic term, so we use lowercase for appliance (according to the PM)
- Start with uppercase for messages or log outputs
- Do not use fullstop at the end of the messages or log outputs

gokhan-berberoglu:
> @gokhan-berberoglu I added a make target for generating the markdowns, so now you can just run `make docs` when you need to update them.

nice 👌 I did not commit them yet to reduce the number of changed files 😅

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/creack/pty v1.1.18
 	github.com/denisbrodbeck/machineid v1.0.1
 	github.com/enriquebris/goconcurrentqueue v0.6.3
+	github.com/gomarkdown/markdown v0.0.0-20221013030248-663e2500819c // indirect
```
danijeel, line 16: You need to run `go mod tidy` once, and commit the changes

gokhan-berberoglu: done

kajes: makes no difference, there's no tagged release for this lib.

danijeel (approved):
LGTM, however we might get some unwanted tech debt by doing Camelcase on all wrapped error messages.
This can lead to strange grammar in the messages to logs/end-user. But its quite minor and can be resolved in a later stage.

PR #351, opened 2022-11-25
sdpctl appliance backup api cmd include parent flag --no-interactive

danijeel:


thomascellerier (approved):
LGTM.

A general sdpctl note: We should really fail the command immediately if passed unknown flags though.
That can result in unexpected behavior, especially if making a typo in e.g. a filter flag.

kajes (approved):

PR #355, opened 2022-11-28
add execution path for non-rebooted appliances

kajes:
Appliances that has gone through an upgrade complete, but failed to reboot will now be accounted for properly.

This will make those appliances skip the actual upgrade complete API request and jump straight to the reboot.

### DIFF
--- a/cmd/appliance/upgrade/complete.go
+++ b/cmd/appliance/upgrade/complete.go
@@ -254,6 +254,8 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
 		}
 	}
 
+	rebootRegex := regexp.MustCompile(`a reboot is required for the upgrade to go into effect`)
+	toReboot := []string{}
 	upgradeStatuses, err := a.UpgradeStatusMap(ctx, appliances)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


cmd/appliance/upgrade/complete.go
```diff
@@ -266,6 +268,8 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
 					appliances = append(appliances[:i], appliances[i+1:]...)
 				}
 			}
+		} else if result.Status == appliancepkg.UpgradeStatusSuccess && rebootRegex.MatchString(result.Details) {
```
thomascellerier, line 271: Why do we need to match the details here?
The success status should be enough to know that we need to issue the upgrade switch partition command.

This couples the logic unnecessarily to the wording of the details IMO.

thomascellerier (approved):
Looks good in general but not sure about matching on details.
It makes sense in ready upgrade status since we use it to get the target version.

Longer term we need to make the API richer on the controller/appliance side to not require that kind of regexing.

cmd/appliance/upgrade/complete.go
kajes: Well, it depends on weather there can really be any other message when the appliance is in a success state? If not, then yes, this is not neccessary.

thomascellerier: No but there is no other purpose for the success state, so why check the message at all?

thomascellerier: This prevents us from rewording if e.g. Jamie finds some better wording from an English ,anguage perspective.

kajes: now it's removed

PR #359, opened 2022-12-16
Return error when command fails to get a configured hostname for the primary controller

kajes:
This fixes an error that occur if the command, for some reason, fails to fetch a configured hostname for the primary controller, rendering it impossible for the appliances to know where to download the upgrade image from.


danijeel (approved):
:+1: Interesting

thomascellerier (approved):
Weird, not sure why that would fail apart from the Controller failing to respond at all.
But better be safe for sure.

PR #366, opened 2023-01-10
Remove made up copyright date

userappgate:


cmd/help.go
```diff
@@ -13,7 +13,7 @@ import (
 func HelpTemplate() string {
 	cobra.AddTemplateFunc("now", time.Now)
 	cobra.AddTemplateFunc("caller", getCaller)
-	return `© {{ now.Year }} Appgate Cybersecurity, Inc.
+	return `© Appgate Cybersecurity, Inc.
```
kajes, line 16:
```suggestion
	return `© 2022-{{ now.Year }} Appgate Cybersecurity, Inc.
```
This is better.

userappgate: it's incorrect. Copyright is about when it was written, not when it's being executed.

cmd/help.go
```diff
@@ -13,7 +13,7 @@ import (
 func HelpTemplate() string {
 	cobra.AddTemplateFunc("now", time.Now)
```
danijeel, line 14:
```suggestion
```

danijeel (approved):
Removed un-used template variable  `cobra.AddTemplateFunc("now", time.Now)`

cmd/help.go
kajes: It's also supposed to guide when the copyright may expire, which it doesn't do if you remove the year entirely. I don't think I've seen a copyright like this without a year provided.

userappgate: How can it guide on when the copyright expires if it always shows today's date?

kajes: Just leave `2022` there, and an indicator that is wrong for a short time every year is absolutely better than nothing at all

kajes: actually, just print 2022-2023. By the time we release the next version, it's going to be correct

Issue #364, opened 2023-01-05
Seeing delay of up to 4-5 minute per interaction


zharif-zakaria:
Was running the latest linux binary release tar.gz, but I was seeing delay of 5 minute per api call. 
Below is the output with log time and ip removed. 

```
username:~$ ./sdpctl configure signin --debug --no-verify --ci-mode
INFO   [2023-01-05T06:43:00Z]                                               COMMAND="sdpctl configure signin" SDPCTL_VERSION=2022.12.20
WARNING[2023-01-05T06:47:00Z] Failed to delete auth token                   error="Failed to activate service 'org.freedesktop.secrets': timed out"
2023/01/05 06:51:01 
POST /admin/authentication HTTP/1.1
Host: XX.XX.XX.XX:8443
User-Agent: sdpctl/2022.12.20
Content-Length: 70
Accept: application/vnd.appgate.peer-v5+json
Content-Type: application/json
Accept-Encoding: gzip
```

PR #368, opened 2023-01-19
Timeout limit subprocess

danijeel:
Set a fixed timeout of  5 seconds on zalando/go-keyring/go-keyring operations, this only effect linux systems.


kajes (approved):
I think this fine for now, but I would like a task where we have fallbacks in cases like these instead.

kajes:
Fixes #364

PR #372, opened 2023-01-27
Fixed typos

aaronpalermo:
Replaced 'mange' with 'manage' in several files.


thomascellerier:
Looks good, you need to sign the CLA first before we can merge.

kajes (changes requested):
Don't merge this now. The html-files are generated, so the typos should be fixed in the source.

kajes (approved):
Ok, just did an extra check, and this seems to change just the things that would be generated anyway, so good to merge :+1: .

PR #374, opened 2023-02-03
read local image file in chunks instead of entire file

danijeel:
During `sdpctl appliance upgrade prepare`, if the user used local image from disk, we read the whole file in memory. We replaced this with read in chunks through io.Pipe to reduce memory footprint.  

This commit resolves SA-21133


kajes (approved):
Huge difference in memory footprint. :+1:

PR #375, opened 2023-02-06
Ability to order appliance lists by keywords

kajes:
Add ability to sort appliance lists, both regular list and stats.

Introduces two new flags, the `--order-by` and `--descending`. The `--order-by` flag accepts a comma seperated list of keywords to sort by in priority. The first word mentioned is going to have higher priority than the next.

The `--descneding` flag is a boolean value which will determine which order to sort. Setting this flag will change the sort order for all provided keywords.

Fixes SA-20828


cmd/appliance/appliance.go
```diff
@@ -27,6 +28,8 @@ func NewApplianceCmd(f *factory.Factory) *cobra.Command {
 	pFlags := cmd.PersistentFlags()
 	pFlags.StringToStringP("include", "i", map[string]string{}, "Include appliances. Adheres to the same syntax and key-value pairs as '--exclude'")
 	pFlags.StringToStringP("exclude", "e", map[string]string{}, filterHelp)
+	pFlags.StringSlice("order-by", []string{"name"}, orderByHelp)
+	pFlags.Bool("descending", false, "Change the direction of sort order when using the '--order-by' flag. Using this will reverse the sort order for all keywords specified in the ''--order-by' flag.")
```
danijeel, line 32: these flags are not applicable for all the appliance commands

testing/helpers.go
```diff
@@ -0,0 +1,88 @@
+package testing
```
danijeel, line 1: this should not be in a new package, these exported methods should be within the existing pkg/appliance, for example pkg/appliance/functions_tests.go where it's only being used.

pkg/appliance/functions.go
```diff
@@ -596,6 +584,134 @@ func applyApplianceFilter(appliances []openapi.Appliance, filter map[string]stri
 	return filteredAppliances, nil
 }
 
+func orderAppliances(appliances []openapi.Appliance, orderBy []string, descending bool) ([]openapi.Appliance, error) {
+	// reverse loop the slice to prioritize the ordering. First entered has priority
+	for i := len(orderBy) - 1; i >= 0; i-- {
+		switch orderBy[i] {
+		case "name":
```
danijeel, line 591: just and idea, but if these key values in the switch case was a exported map instead, it would be trivial to later add support for tab completion in the command that can use these.

pkg/appliance/functions.go
```diff
@@ -596,6 +584,134 @@ func applyApplianceFilter(appliances []openapi.Appliance, filter map[string]stri
...
+				sort.SliceStable(appliances, func(i, j int) bool {
+					return !appliances[i].GetActivated() && appliances[i].GetActivated() != appliances[j].GetActivated()
+				})
+			} else {
+				sort.SliceStable(appliances, func(i, j int) bool {
+					return appliances[i].GetActivated() && appliances[i].GetActivated() != appliances[j].GetActivated()
+				})
+			}
+		default:
+			log.WithField("keyword", orderBy[i]).Warn("keyword not sortable, ignoring")
```
danijeel, line 632: redundant log, if you really want log for these, maybe we should demote it to DEBUG?

pkg/appliance/functions.go
```diff
@@ -596,6 +584,134 @@ func applyApplianceFilter(appliances []openapi.Appliance, filter map[string]stri
...
+				sort.SliceStable(stats, func(i, j int) bool { return stats[i].GetFunction() < stats[j].GetFunction() })
+			}
+		case "status":
+			if descending {
+				sort.SliceStable(stats, func(i, j int) bool { return stats[i].GetStatus() > stats[j].GetStatus() })
+			} else {
+				sort.SliceStable(stats, func(i, j int) bool { return stats[i].GetStatus() < stats[j].GetStatus() })
+			}
+		default:
+			log.WithField("keyword", orderBy[i]).Warn("not a sortable keyword")
```
danijeel, line 709: redundant log, if you really want log for these, maybe we should demote it to DEBUG?

pkg/appliance/functions.go
kajes: a map is not ordered though, so it would be impossible to implement prioritized ordering by using a map.

pkg/appliance/functions.go
kajes: I wouldn't call any of these redundant. If I typo one of the kywords and don't understand why the ordering isn't the way it's supposed to be, I wouldn't want to need to set up DEBUG mode just to figure out that I mistyped something.

testing/helpers.go
```diff
@@ -0,0 +1 @@
...
```
danijeel, line 1: `rm -rf testing` this file and directory is not needed.

cmd/appliance/appliance.go
thomascellerier: Could that code be reused for the relevant commands instead?
E.g. some kind of "arg group" that you can pass to the appropriate subcommands.

kajes: Not that I know of, the easiest way would be to hide from the commands where it's not applicable. On the other hand, there are very few appliance commands that doesn't list appliances at all. I'm not sure I understand really which appliance commands this wouldn't be applicable to?

danijeel: wouldn't it be easier to just have these flags in stats and appliance list command where there are used?

pkg/appliance/functions.go
```diff
@@ -596,6 +584,134 @@ func applyApplianceFilter(appliances []openapi.Appliance, filter map[string]stri
...
+			} else {
+				sort.SliceStable(appliances, func(i, j int) bool {
+					return appliances[i].GetActivated() && appliances[i].GetActivated() != appliances[j].GetActivated()
+				})
+			}
+		default:
+			log.WithField("keyword", orderBy[i]).Warn("keyword not sortable, ignoring")
+		}
+	}
+	return appliances, nil
```
danijeel, line 635: instead of doing all these  `if descending { /**/ } else { /**/ }`
couldn't we reverse the slice instead?
```go
	if descending {
		return reverse(appliances), nil
	}
	return appliances, nil
```

cmd/appliance/appliance.go
kajes: There are lists being used in basically every appliance command. If it doesn't output a list, it will prompt with a list if you don't use arguments, or output the list in a summary. Even when prompting, being able to order by specific values is useful.

pkg/appliance/functions.go
kajes: Good point. Fixed :+1:

pkg/appliance/functions.go
```diff
@@ -596,6 +584,86 @@ func applyApplianceFilter(appliances []openapi.Appliance, filter map[string]stri
 	return filteredAppliances, nil
 }
 
+func reverse[T openapi.Appliance | openapi.StatsAppliancesListAllOfData](a []T) []T {
+	rev := []T{}
+	for i := len(a) - 1; i >= 0; i-- {
+		rev = append(rev, a[i])
+	}
+	return rev
+}
```
danijeel, line 593: just nitpicking now, but we can make it even more generic
```suggestion
func reverse[S ~[]T, T any](items S) S {
	if len(items) <= 1 {
		return items
	}
	result := make([]T, 0, len(items))
	for i := len(items) - 1; i >= 0; i-- {
		result = append(result, items[i])
	}
	return result
}
```

kajes: Sure. Still not used to the generics, but this could definitely be more generic.

cmd/appliance/appliance.go
thomascellerier: Do you mean even including inside the prepare upgrade etc.. this will sort appliances according to the flag?

pkg/appliance/functions.go
thomascellerier: In that case I think the command should fail with a non zero exit code immediately. No point in running at all if the command can't do what we tell it to do.

cmd/appliance/appliance.go
kajes: yes, that's what I'm saying. Complete has it's own logic for grouping the appliances, but the flags will affect how the appliances are listed within the groups

thomascellerier (approved):
Having these as generic flags seems reasonable to me given that the ordering is used when displaying prompts as well.

Approving on functionality, trusting @danijeel 's review for the code part.

kajes:
Tyron is ok with having it as a general flag for appliance. For now, I think it's fine to merge, but we'll probably re-visit list sorting soon again. There's still list unification to be done.

Any other comments on code @danijeel ?

danijeel (approved):

PR #377, opened 2023-02-09
nil err in api caused panic on String()

danijeel:
found this while testing other stuff, got panic on nil err when using toString method.


kajes (approved):
LGTM

PR #378, opened 2023-02-10
periodic version checking

kajes:
Check for a new version of sdpctl when running commands. If a new version is found, a notice will be printed at the top of the command output with a link where the new version can be downloaded.

Version is checked periodically, once every day, but can optionally be disabled in the config file.

Fixes: SA-20202

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -273,11 +273,6 @@ func rootPersistentPreRunEFunc(f *factory.Factory, cfg *configuration.Config) fu
 				if err == nil && minMax != nil {
 					viper.Set("api_version", minMax.Max)
 					f.Config.Version = int(minMax.Max)
-					if err := viper.WriteConfig(); err != nil {
-						if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
-							fmt.Fprintf(f.StdErr, "[error] %s\n", err)
-						}
-					}
 				}
 			}
 
@@ -301,13 +296,19 @@ func rootPersistentPreRunEFunc(f *factory.Factory, cfg *configuration.Config) fu
 		}
 
 		// Check for new sdpctl version
-		meta, err := cfg.CheckForUpdate(f.StdErr, version)
+		client, err := f.HTTPClient()
+		if err != nil {
+			return err
+		}
+		meta, err := cfg.CheckForUpdate(f.StdErr, client, version)
 		if err != nil {
 			log.WithError(err).Info("version check result")
 		}
 		viper.Set("meta", meta)
 		if err := viper.WriteConfig(); err != nil {
-			return err
+			if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
+				fmt.Fprintf(f.StdErr, "[error] %s\n", err)
+			}
 		}
 
 		return nil
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -300,11 +300,10 @@ func rootPersistentPreRunEFunc(f *factory.Factory, cfg *configuration.Config) fu
 		if err != nil {
 			return err
 		}
-		meta, err := cfg.CheckForUpdate(f.StdErr, client, version)
+		cfg, err = cfg.CheckForUpdate(f.StdErr, client, version)
 		if err != nil {
 			log.WithError(err).Info("version check result")
 		}
-		viper.Set("meta", meta)
 		if err := viper.WriteConfig(); err != nil {
 			if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
 				fmt.Fprintf(f.StdErr, "[error] %s\n", err)
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -301,8 +301,13 @@ func rootPersistentPreRunEFunc(f *factory.Factory, cfg *configuration.Config) fu
 		}
 
 		// Check for new sdpctl version
-		if err := cfg.CheckForUpdate(f.IOOutWriter, version); err != nil {
+		meta, err := cfg.CheckForUpdate(f.StdErr, version)
+		if err != nil {
 			log.WithError(err).Info("version check result")
+		}
+		viper.Set("meta", meta)
+		if err := viper.WriteConfig(); err != nil {
+			return err
 		}
 
 		return nil
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -33,7 +33,7 @@ import (
 )
 
 var (
-	version         string = "dev"
+	version         string = "0.0.0-dev"
 	commit          string
 	buildDate       string
 	longDescription string = `The official CLI tool for managing your Collective.`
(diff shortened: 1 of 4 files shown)


pkg/configuration/configuration.go
```diff
@@ -280,3 +290,74 @@ func (c *Config) KeyringPrefix() (string, error) {
...
+
+	// Perform version check
+	const cliReleasesURL = "https://api.github.com/repos/appgate/sdpctl/releases"
+	req, err := http.NewRequest(http.MethodGet, cliReleasesURL, nil)
+	if err != nil {
+		return err
+	}
+	req.Header.Add("Accept", "application/vnd.github+json")
+	client := &http.Client{}
+	res, err := client.Do(req)
```
danijeel, line 315: we need to set a timeout here, it's very likely that github.com can be blocked/unreachable for the customers

pkg/configuration/configuration.go
```diff
@@ -280,3 +290,74 @@ func (c *Config) KeyringPrefix() (string, error) {
...
+
+func (c *Config) CheckForUpdate(out io.Writer, current string) error {
+	// Check if version check is disabled in configuration
+	if c.Meta.DisableVersionCheck {
+		return errors.New("version check disabled")
+	}
+
+	// Check if version check has already been done today
+	yesterday := time.Now().AddDate(0, 0, -1)
+	lastCheck, _ := time.Parse(time.RFC3339Nano, c.Meta.LastChecked)
```
danijeel, line 302: dont supress errors

cmd/root.go
```diff
@@ -299,6 +299,12 @@ func rootPersistentPreRunEFunc(f *factory.Factory, cfg *configuration.Config) fu
 		if err := configuration.CheckMinAPIVersionRestriction(cmd, cfg.Version); err != nil {
 			return err
 		}
+
+		// Check for new sdpctl version
+		if err := cfg.CheckForUpdate(f.IOOutWriter, version); err != nil {
```
danijeel, line 304: should probably be stderr instead so we dont block pipe operations

pkg/configuration/configuration.go
```diff
@@ -280,3 +290,74 @@ func (c *Config) KeyringPrefix() (string, error) {
 	}
 	return h, nil
 }
+
+func (c *Config) CheckForUpdate(out io.Writer, current string) error {
```
danijeel, line 294: I would suggest return Meta and write it in root, and catch write errors as done in the other cases
```suggestion
func (c *Config) CheckForUpdate(out io.Writer, current string) (Meta,error) {
```

pkg/configuration/configuration.go
kajes: We should probably write a new last check time too, even though the request fails, so we don't retry on every consecutive command when Github is unreachable.

pkg/configuration/configuration.go
```diff
@@ -280,3 +291,76 @@ func (c *Config) KeyringPrefix() (string, error) {
...
+
+	// Check if version check has already been done today
+	yesterday := time.Now().AddDate(0, 0, -1)
+	lastCheck, err := time.Parse(time.RFC3339Nano, c.Meta.LastChecked)
+	if err != nil {
+		return c.Meta, err
+	}
+	if !lastCheck.Before(yesterday) {
+		return c.Meta, errors.New("version check already done today")
+	}
```
danijeel, line 309:
```suggestion
	// Check if version check has already been done today
	lastCheck, err := time.Parse(time.RFC3339Nano, c.Meta.LastChecked)
	if err == nil {
        yesterday := time.Now().AddDate(0, 0, -1)
		if !lastCheck.Before(yesterday) {
			return c.Meta, errors.New("version check already done today")
		}
	}
```

otherwise it will always fail here if configure through env variables without any config file

cmd/root.go
```diff
@@ -299,6 +299,17 @@ func rootPersistentPreRunEFunc(f *factory.Factory, cfg *configuration.Config) fu
...
 			return err
 		}
+
+		// Check for new sdpctl version
+		meta, err := cfg.CheckForUpdate(f.StdErr, version)
+		if err != nil {
+			log.WithError(err).Info("version check result")
+		}
+		viper.Set("meta", meta)
+		if err := viper.WriteConfig(); err != nil {
```
danijeel, line 309: refactor with https://github.com/appgate/sdpctl/blob/23e58b1d42673a64e6a4b9632b215895e25a73a1/cmd/root.go#L276-L280 so we only write once, and handle the case if we configure sdpctl with environment variables only without config file. I'm pretty sure the current version here would fail in in the e2e test

pkg/configuration/configuration.go
```diff
@@ -280,3 +291,76 @@ func (c *Config) KeyringPrefix() (string, error) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, cliReleasesURL, nil)
+	// Write new check time to config after request is made
+	c.Meta.LastChecked = time.Now().Format(time.RFC3339Nano)
+	if err != nil {
+		return c.Meta, err
+	}
+	req.Header.Add("Accept", "application/vnd.github+json")
+	client := &http.Client{}
```
danijeel, line 322: might be worth using the HTTPClient from the factory https://github.com/appgate/sdpctl/blob/23e58b1d42673a64e6a4b9632b215895e25a73a1/pkg/factory/http.go#L26 that we we respect HTTP_PROXY for all http requests within the project.

pkg/configuration/configuration.go
```diff
@@ -280,3 +291,76 @@ func (c *Config) KeyringPrefix() (string, error) {
...
+
+	type githubRelease struct {
+		TagName    string `json:"tag_name"`
+		PreRelease bool   `json:"pre_release"`
+		Draft      bool   `json:"draft"`
+	}
+
+	type releaseList []githubRelease
+
+	b, err := io.ReadAll(res.Body)
```
danijeel, line 337: we should probably check HTTP Status  code is 200 before continuing here

pkg/configuration/configuration.go
kajes: Yeah, I don't see any issues with it either.

pkg/configuration/configuration.go
```diff
@@ -28,6 +38,7 @@ type Config struct {
 	DeviceID    string  `mapstructure:"device_id"`
 	PemFilePath string  `mapstructure:"pem_filepath"`
 	Timeout     int     // HTTP timeout, not supported in the config file.
+	Meta        Meta    `mapstructure:"meta"`
```
danijeel, line 41: I think we should keep a flat config structure. Right now there is no way to disable version check through environment variables. but If we would use a flat config, viper would take care of it for us.

kajes: Good point. I'll fix that as well

danijeel (approved):
if the e2e passes, we can merge this.

PR #382, opened 2023-02-22
Fix deprecated yaml values in new goreleaser versions

kajes:
change deprecated yaml value for goreleaser


danijeel (approved):
What is the deprecation notice message? ❓

kajes:
> What is the deprecation notice message? question

Couldn't find any actually on the goreleaser page, but it wouldn't build with the documentation values.

PR #383, opened 2023-02-24
Token list sorting

kajes:
Implements list sorting for the `tokens list` command. Similar to the appliance list sorting, it uses the `--order-by` and `--descending` flags, where the former accepts a comma separated list of sortable keywords, while the latter reverses the sort order if set.

Also included are some minor build issues found:
* add missing docs for the `maintenance status`
* set correct permissions for included files in deb/rpm package

### DIFF
--- a/cmd/token/token.go
+++ b/cmd/token/token.go
@@ -28,10 +28,9 @@ func NewTokenCmd(f *factory.Factory) *cobra.Command {
 	}
 
 	var tokenCmd = &cobra.Command{
-		Use:     "token",
-		Aliases: []string{"tokens"},
-		Short:   "Perform actions on Admin, Claims and Entitlement tokens",
-		Long:    `The token command allows you to renew or revoke tokens used in the Collective.`,
+		Use:   "token",
+		Short: "Perform actions on Admin, Claims and Entitlement tokens",
+		Long:  `The token command allows you to renew or revoke tokens used in the Collective.`,
 	}
 
 	tokenCmd.PersistentFlags().BoolVar(&opts.useJSON, "json", false, "Display in JSON format")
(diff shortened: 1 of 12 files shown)


pkg/docs/maintenance.go
```diff
@@ -15,6 +15,11 @@ var (
 		Long:  ``,
 	}
 
+	MaintenanceStatusDoc = CommandDoc{
+		Short: "View maintenance status of appliances",
+		Long:  "",
```
kosmas-valianos, line 20: That's a `string` I guess then default value is already `""`

kosmas-valianos (commented):
LGTM but for logic checks better @danijeel to check as well

cmd/token/token.go
```diff
@@ -26,9 +28,10 @@ func NewTokenCmd(f *factory.Factory) *cobra.Command {
 	}
 
 	var tokenCmd = &cobra.Command{
-		Use:   "token",
-		Short: "Perform actions on Admin, Claims and Entitlement tokens",
-		Long:  `The token command allows you to renew or revoke tokens used in the Collective.`,
+		Use:     "token",
+		Aliases: []string{"tokens"},
```
danijeel, line 32: Let's keep it consistent with all the other commands, and not put plural on our noun commands.

danijeel:
since this PR includes Debian fixes, I will assign @userappgate to this PR

userappgate:
You expect me to look at a completely non standard and having nothing to do with debian `.goreleaser.yml` file and understand if it's correct or not? :D

userappgate (approved):
It seems to install fine and not have man pages listed as config files.

PR #387, opened 2023-03-13
Prevent same site gateways from batching together when upgrading

kajes:
This fixes a bug where same site gateways were sometimes batched together when upgrading, causing downtime for the site when completing an upgrade.

Gateways are now treated specially if they are enabled, giving them the same hash code if they belong to the same site and are enabled. An additional check is also made to prevent enabled gateways to be batched together with same site gateways when re-balancing of batches occurs. It is better to have a batch with only one appliance rather than causing downtime.

Fixes SA-21318

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,7 @@
 root = true
 
 [*]
-indent_style = space
+indent_style = tab
 indent_size = 4
 end_of_line = lf
 charset = utf-8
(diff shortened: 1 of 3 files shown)


danijeel (approved):

thomascellerier (approved):
Looks good, the same logic needs to be applied to connectors on the same site.

kajes:
I'll add the connector logic as well before merging. Very similar, so won't be a big change

PR #389, opened 2023-03-15
Better information on disk usage

kajes:
This implements a more detailed view of disk usage in `appliance stats` and in the disk space warning.

It uses the new `diskInfo` object provided by the appliance API since v18. If a lower version than v18 is used, the old disk info is shown instead.

Also contains a small change for the list ordering of appliances. Sorting by `site` now orders the list by site name (just as using `site-name` would) while ordering by site ID will require using the specific `site-id` keyword instead.

Fixes SA-19775

### DIFF
--- a/cmd/appliance/stats.go
+++ b/cmd/appliance/stats.go
@@ -84,7 +84,7 @@ func statsRun(cmd *cobra.Command, args []string, opts *statsOptions) error {
 			fmt.Sprintf("%g%%", s.GetCpu()),
 			fmt.Sprintf("%g%%", s.GetMemory()),
 			statsNetworkPrettyPrint(s.GetNetwork()),
-			statsDiskUsage(cfg.Version, s),
+			statsDiskUsage(s),
 			version,
 		)
 	}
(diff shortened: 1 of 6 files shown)


thomascellerier (approved):

cmd/appliance/stats.go
```diff
@@ -92,6 +96,17 @@ func statsNetworkPrettyPrint(n openapi.StatsAppliancesListAllOfNetwork) string {
...
+func statsDiskUsage(apiVersion int, stats openapi.StatsAppliancesListAllOfData) string {
+	if apiVersion < 18 {
+		return fmt.Sprintf("%g%%", stats.GetDisk())
+	}
+	diskInfo := stats.GetDiskInfo()
+	used, total := diskInfo.GetUsed(), diskInfo.GetTotal()
+	percentUsed := (used / total) * 100
+	r := fmt.Sprintf("%.2f%% (%s / %s)", percentUsed, appliancepkg.PrettyBytes(float64(used)), appliancepkg.PrettyBytes(float64(total)))
+	return r
+}
```
danijeel, line 108: :pick: You could remove apiVersion int here and reduce complexity by reversing it
```suggestion
func statsDiskUsage(stats openapi.StatsAppliancesListAllOfData) string {
	if v, ok := stats.GetDiskInfoOk(); ok {
		diskInfo := *v
		used, total := diskInfo.GetUsed(), diskInfo.GetTotal()
		percentUsed := (used / total) * 100
		return fmt.Sprintf("%.2f%% (%s / %s)", percentUsed, appliancepkg.PrettyBytes(float64(used)), appliancepkg.PrettyBytes(float64(total)))
	}
	return fmt.Sprintf("%g%%", stats.GetDisk())
}
```

pkg/appliance/checks.go
```diff
@@ -15,11 +15,22 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-func PrintDiskSpaceWarningMessage(out io.Writer, stats []openapi.StatsAppliancesListAllOfData) {
+func PrintDiskSpaceWarningMessage(out io.Writer, stats []openapi.StatsAppliancesListAllOfData, apiVersion int) {
```
danijeel, line 18: :pick: same here, you could remove apiVersion check here and only check if we have DiskInfo, otherwise fallback to DiskUsage

cmd/appliance/stats.go
kajes: :+1:

pkg/appliance/checks.go
kajes: The table header is a problem here though. Appliance stats are in a slice and is looped to add lines to the table. The table header though is set before looping the appliance stats slice, so `GetDiskInfoOk` is not available at that point, and removing the `apiVersion` param would make it even more complicated.

danijeel: could be solved by simple sharing the title between the different values, for example `Disk` or `Disk usage` and the user can deduce what it is based on the unit (percentage or megabyte).

kajes: I'm pretty sure we are going to get confused users if we do that though. I'd rather have the code a little bit more complex than risk the confusion.

PR #394, opened 2023-04-04
LogServer image bundling

kajes:
From appliance version 6.2, the LogServer function is decoupled from the base appliance image. Upgrading LogServer appliances will be done by downloading the required docker images for LogServer and uploading it to the Primary Controller, just like the base image is being uploaded already.

This PR will automatically pull the required LogServer images needed, pack them into a zip-file and upload it when running `upgrade prepare`.

Fixes: SA-21477

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -20,6 +20,9 @@ builds:
         - -s -w -X "github.com/appgate/sdpctl/cmd.version={{ if .Tag }}{{ .Tag }}{{ else }}dev{{ end }}"
         - -X "github.com/appgate/sdpctl/cmd.commit={{ .FullCommit }}"
         - -X "github.com/appgate/sdpctl/cmd.buildDate={{ .Date }}"
+        - -X "github.com/appgate/sdpctl/cmd/appliance/upgrade.dockerRegistry={{ .Env.DOCKER_REGISTRY_URL }}"
+        - -X "github.com/appgate/sdpctl/cmd/appliance/upgrade.dockerRegistryUsername={{ .Env.DOCKER_REGISTRY_USERNAME }}"
+        - -X "github.com/appgate/sdpctl/cmd/appliance/upgrade.dockerRegistryPassword={{ .Env.DOCKER_REGISTRY_PASSWORD }}"
     id: linux
     goos: [linux]
     goarch: [amd64, arm64]
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -21,8 +21,6 @@ builds:
         - -X "github.com/appgate/sdpctl/cmd.commit={{ .FullCommit }}"
         - -X "github.com/appgate/sdpctl/cmd.buildDate={{ .Date }}"
         - -X "github.com/appgate/sdpctl/cmd/appliance/upgrade.dockerRegistry={{ .Env.DOCKER_REGISTRY_URL }}"
-        - -X "github.com/appgate/sdpctl/cmd/appliance/upgrade.dockerRegistryUsername={{ .Env.DOCKER_REGISTRY_USERNAME }}"
-        - -X "github.com/appgate/sdpctl/cmd/appliance/upgrade.dockerRegistryPassword={{ .Env.DOCKER_REGISTRY_PASSWORD }}"
     id: linux
     goos: [linux]
     goarch: [amd64, arm64]
--- a/cmd/appliance/upgrade/prepare.go
+++ b/cmd/appliance/upgrade/prepare.go
@@ -38,6 +38,8 @@ import (
 	"github.com/vbauerster/mpb/v7/decor"
 )
 
+var dockerRegistry string
+
 type prepareUpgradeOptions struct {
 	Config           *configuration.Config
 	Out              io.Writer
@@ -56,6 +58,7 @@ type prepareUpgradeOptions struct {
 	ciMode           bool
 	actualHostname   string
 	targetVersion    *version.Version
+	dockerRegistry   string
 }
 
 // NewPrepareUpgradeCmd return a new prepare upgrade command
@@ -101,6 +104,23 @@ func NewPrepareUpgradeCmd(f *factory.Factory) *cobra.Command {
 			if err := checkImageFilename(opts.filename); err != nil {
 				return err
 			}
+
+			// Get the docker registry address
+			opts.dockerRegistry = os.Getenv("SDPCTL_DOCKER_REGISTRY")
+			if len(opts.dockerRegistry) <= 0 {
+				opts.dockerRegistry, err = cmd.Flags().GetString("docker-registry")
+				if err != nil {
+					return err
+				}
+				if len(opts.dockerRegistry) <= 0 {
+					if len(dockerRegistry) > 0 {
+						opts.dockerRegistry = dockerRegistry
+					} else {
+						return errors.New("no docker registry address found")
+					}
+				}
+			}
+			log.WithField("URL", opts.dockerRegistry).Debug("found docker registry address")
 
 			// allow remote addr for image, such as aws s3 bucket
 			if util.IsValidURL(opts.image) {
@@ -169,6 +189,7 @@ func NewPrepareUpgradeCmd(f *factory.Factory) *cobra.Command {
 	flags.BoolVar(&opts.hostOnController, "host-on-controller", false, "Use the primary Controller as image host when uploading from remote source")
 	flags.StringVar(&opts.actualHostname, "actual-hostname", "", "If the actual hostname is different from that which you are connecting to the appliance admin API, this flag can be used for setting the actual hostname")
 	flags.BoolVar(&opts.forcePrepare, "force", false, "Force prepare of upgrade on appliances even though the version uploaded is the same or lower than the version already running on the appliance")
+	flags.StringVar(&opts.dockerRegistry, "docker-registry", "", "Custom registry where the docker images should be fetched from")
 
 	return prepareCmd
 }
--- a/cmd/appliance/upgrade/prepare.go
+++ b/cmd/appliance/upgrade/prepare.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"mime/multipart"
+	"net/http"
 	"net/url"
 	"os"
 	"path"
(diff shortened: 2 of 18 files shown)


pkg/factory/http.go
```diff
@@ -163,6 +166,51 @@ func customHTTPClient(f *Factory) func() (*http.Client, error) {
...
+			underlyingTransport: parentTransport,
+			username:            username,
+			password:            password,
+			useragent:           f.userAgent,
+			accept:              fmt.Sprintf("application/vnd.appgate.peer-v%d+json", f.Config.Version),
+		}
+		return client, nil
+	}
+}
+
```
danijeel, line 213: Why is this needed? wouldn't it be enough to simply add the Authorization header key, value on the request?

.goreleaser.yml
```diff
@@ -20,6 +20,9 @@ builds:
         - -s -w -X "github.com/appgate/sdpctl/cmd.version={{ if .Tag }}{{ .Tag }}{{ else }}dev{{ end }}"
         - -X "github.com/appgate/sdpctl/cmd.commit={{ .FullCommit }}"
         - -X "github.com/appgate/sdpctl/cmd.buildDate={{ .Date }}"
+        - -X "github.com/appgate/sdpctl/cmd/appliance/upgrade.dockerRegistry={{ .Env.DOCKER_REGISTRY_URL }}"
+        - -X "github.com/appgate/sdpctl/cmd/appliance/upgrade.dockerRegistryUsername={{ .Env.DOCKER_REGISTRY_USERNAME }}"
+        - -X "github.com/appgate/sdpctl/cmd/appliance/upgrade.dockerRegistryPassword={{ .Env.DOCKER_REGISTRY_PASSWORD }}"
```
danijeel, line 25: what is the reasoning behind this? why is these build vars? just curious...

pkg/appliance/functions.go
```diff
@@ -776,3 +788,251 @@ func PrettyBytes(v float64) string {
...
+	manifestReq, err := http.NewRequestWithContext(ctx, http.MethodGet, manifestURL, nil)
+	if err != nil {
+		mainErrChan <- err
+		return
+	}
+	for k, v := range headers {
+		manifestReq.Header.Set(k, v)
+	}
+	manifestRes, err := client.Do(manifestReq)
+	if manifestRes.StatusCode != http.StatusOK || err != nil {
```
danijeel, line 869: manifestRes can be null, should be enough to only check err
```suggestion
	if err != nil {
```

pkg/appliance/functions.go
```diff
@@ -776,3 +788,251 @@ func PrettyBytes(v float64) string {
...
+	configReq, err := http.NewRequestWithContext(ctx, http.MethodGet, configURL, nil)
+	if err != nil {
+		mainErrChan <- err
+		return
+	}
+	for k, v := range headers {
+		configReq.Header.Set(k, v)
+	}
+	configRes, err := client.Do(configReq)
+	if configRes.StatusCode != http.StatusOK || err != nil {
```
danijeel, line 898:
```suggestion
	if err != nil {
```
configRes can be null, should be enough to only check err

pkg/appliance/functions.go
```diff
@@ -776,3 +788,251 @@ func PrettyBytes(v float64) string {
...
+		go func(l ManifestConfig, bufChan chan bufName, errChan chan error) {
+			f := strings.Replace(l.Digest, "sha256:", "", 1) + ".tar.gz"
+			layerURL := fmt.Sprintf("%s/v2/%s/blobs/%s", registry, image, l.Digest)
+			layerReq, err := http.NewRequestWithContext(ctx, http.MethodGet, layerURL, nil)
+			if err != nil {
+				errChan <- err
+				return
+			}
+			layerRes, err := client.Do(layerReq)
+			if layerRes.StatusCode != http.StatusOK || err != nil {
```
danijeel, line 929:
```suggestion
			if err != nil {
```
layerRes can be null, avoid panic by only checking err

pkg/appliance/functions.go
```diff
@@ -776,3 +788,251 @@ func PrettyBytes(v float64) string {
...
+		if err != nil {
+			errs = multierror.Append(errs, err)
+		}
+	}
+	log.Info("docker layers downloaded")
+
+	return archive, fileStat, errs.ErrorOrNil()
+}
+
+func downloadDockerImageBundle(ctx context.Context, out io.Writer, client *http.Client, ciMode bool, zipWriter *zip.Writer, registry, image, tag string, p *tui.Progress, mainErrChan chan<- error, wg *sync.WaitGroup) {
```
danijeel, line 851: some inline code comments should be suitable for this func, with referance to docker docs.

cmd/appliance/upgrade/prepare.go
```diff
@@ -34,14 +35,23 @@ import (
 	"github.com/hashicorp/go-version"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-	"github.com/vbauerster/mpb/v7"
-	"github.com/vbauerster/mpb/v7/decor"
+)
+
+var (
+	// Set at build time or in env varibales
```
danijeel, line 41:
```suggestion
	// Set at build time or in environment variables
```

pkg/appliance/functions.go
```diff
@@ -776,3 +788,251 @@ func PrettyBytes(v float64) string {
...
+	Layers        []ManifestConfig
+}
+
+type ImageJSON struct {
+	Image string `json:"image"`
+}
+
+func DownloadDockerBundles(ctx context.Context, out io.Writer, client *http.Client, zipName, registry string, images map[string]string, ciMode bool) (*os.File, fs.FileInfo, error) {
+	// Create zip-archive
+	archive, err := os.Create(fmt.Sprintf("/tmp/%s.zip", zipName))
```
danijeel, line 810: this wont work on windows, use https://pkg.go.dev/os#CreateTemp instead

pkg/appliance/functions.go
```diff
@@ -776,3 +788,251 @@ func PrettyBytes(v float64) string {
...
+	}
+
+	log.Info("downloading image layers for ", zipName)
+	var p *tui.Progress
+	if !ciMode {
+		p = tui.New(ctx, out)
+		defer p.Wait()
+	}
+	errChan := make(chan error)
+	var wg sync.WaitGroup
```
danijeel, line 829: https://pkg.go.dev/golang.org/x/sync/errgroup could work here as a option, that way we get rid of errChan, since we don't want to continue if any of these downloads gets  error. that would reduce complexity and func signature

pkg/factory/http.go
kajes: May be a bit overworked here, but it's mainly to make the check for authorization a bit cleaner in the prepare command, but now that I look at it, there's a bit of double checking the env variables too, so will fix that. Anyway, there's quite a lot of requests being made against the registry, so not having to deal with setting a header at each request was a conveniance, but I can simplify it if that's preferred.

pkg/appliance/functions.go
kajes: `401 Unauthorized` does not result in error though, which is a bit important if we want to support basic auth. Though the check should probably be reversed here if configRes can be null, so err should probably be checked first.

pkg/appliance/functions.go
kajes: same as configRes here

pkg/appliance/functions.go
kajes: :+1:

pkg/appliance/functions.go
danijeel: its still prone for panic, since configRes can be nil, you need to add a nested  if statement if you want to handle custom  statusCode check.

thomascellerier:
I will do some manual testing on this and come back to you with some feedback.

thomascellerier:
Now the bundling and uploading is working as expected.

We still need to validate that the contents of the bundle are in the format we expect. Which we can do after fixing some appliance side issues.

cmd/appliance/upgrade/prepare.go
```diff
@@ -367,8 +408,98 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions)
 		}
 	}
 
+	uploadWithProgress := func(ctx context.Context, reader io.Reader, name string, size int64, headers map[string]string) error {
+		p := tui.New(ctx, spinnerOut)
+		defer p.Wait()
+		endMsg := "uploaded"
+		proxyReader, t := p.FileUploadProgress(name, endMsg, size, reader)
```
danijeel, line 415: looks like you are missing `proxyReader.Close()`

cmd/appliance/upgrade/prepare.go
```diff
@@ -731,8 +819,9 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions)
 
 const prepareUpgradeMessage = `PREPARE SUMMARY
 
-1. Upload upgrade image {{.Filepath}} to Controller
-2. Prepare upgrade on the following appliances:
+1. Bundle and upload LogServer docker image if needed
```
danijeel, line 822: Should this always be shown, even though there are on log server in the collective?

thomascellerier: No this should only be shown if there is a log server being upgraded.
We should make sure the whole flow is only triggered in that case.

cmd/appliance/upgrade/prepare.go
kajes: Yep, I'll fix that :+1:

cmd/appliance/upgrade/prepare.go
kajes: Probably not, but takes a bit of work to avoid it, so I guess the question would be if it's a deal breaker before merging?

thomascellerier: I think so yes, this will confuse people.
It also needs to be only shown for upgrades to 6.2.0+ (with a logserver being upgraded).

kajes: Ok, I'll give it some time then before merge

cmd/appliance/upgrade/prepare.go
```diff
@@ -367,8 +408,98 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions)
...
+				err = uploadWithProgress(ctx, pr, "uploading "+logServerZipName, zipInfo.Size(), headers)
+			} else {
+				err = a.UploadFile(ctx, pr, headers)
+				fmt.Fprint(opts.Out, "Image bundles prepared\n\n")
+			}
+			if err != nil {
+				return err
+			}
+			os.Remove(zipPath)
+			fmt.Fprintln(opts.Out)
```
danijeel, line 496:
```suggestion
```

thomascellerier (approved):

PR #398, opened 2023-04-17
add support for NO_PROXY

danijeel:
this commit difference between HTTP_PROXY and HTTPS_PROXY so technically it breaks backwards comparability


kajes:
Looks fine, but do we need to do anything for the backwards compatability?

kajes (approved):

PR #403, opened 2023-04-20
Use image tag derived from upgrade version

kajes:
Pulling docker image will try to pull an image tag derived from the upgrade version.

Tag is overridable by setting the `SDPCTL_DOCKER_TAG` environment variable when running the command.

Fixes: SA-21648

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -168,11 +168,14 @@ func ApplianceVersionString(v *version.Version) string {
 	return fmt.Sprintf("%d.%d.%d-%s-%s", segments[0], segments[1], segments[2], v.Metadata(), preString)
 }
 
-func DockerTagVersion(v *version.Version) string {
+func DockerTagVersion(v *version.Version) (string, error) {
 	if envTag := os.Getenv("SDPCTL_DOCKER_TAG"); len(envTag) > 0 {
-		return envTag
+		return envTag, nil
+	}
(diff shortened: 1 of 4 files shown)


thomascellerier (approved):
Looks good!

pkg/util/util.go
```diff
@@ -167,3 +167,12 @@ func ApplianceVersionString(v *version.Version) string {
 	}
 	return fmt.Sprintf("%d.%d.%d-%s-%s", segments[0], segments[1], segments[2], v.Metadata(), preString)
 }
+
+func DockerTagVersion(v *version.Version) string {
```
danijeel, line 171: Can we add a sanity check here if v is nil?

pkg/util/util_test.go
```diff
@@ -151,3 +154,48 @@ func TestSearchSlice(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			v, err := version.NewVersion(tt.v)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if len(tt.env) > 0 {
+				os.Setenv("SDPCTL_DOCKER_TAG", tt.env)
+			}
```
danijeel, line 195:
```suggestion
			t.Setenv("SDPCTL_DOCKER_TAG", tt.env)
```

PR #404, opened 2023-04-27
add authorization token for public ecr

kajes:
This will download and add the token needed for bundling docker images when using a public ecr registry.

### DIFF
--- a/pkg/appliance/functions.go
+++ b/pkg/appliance/functions.go
@@ -902,6 +902,30 @@ func downloadDockerImageBundle(args imageBundleArgs) {
 		"Accept":       "application/vnd.docker.distribution.manifest.v2+json",
 		"Content-Type": "application/json",
 	}
+
+	// Get the public token if public ECR registry is used
+	if args.registry.Host == "public.ecr.aws" {
+		type ecrToken struct {
+			Token string
+		}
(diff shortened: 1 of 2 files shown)


pkg/appliance/functions.go
```diff
@@ -902,6 +902,30 @@ func downloadDockerImageBundle(args imageBundleArgs) {
 		"Accept":       "application/vnd.docker.distribution.manifest.v2+json",
 		"Content-Type": "application/json",
 	}
+
+	// Get the public token if public ECR registry is used
+	if args.registry.Host == "public.ecr.aws" {
+		type ecrToken struct {
+			Token string
+		}
+		res, err := args.client.Get("https://public.ecr.aws/token/")
```
danijeel, line 911: docker appends scope in the request as parameters, for example `https://public.ecr.aws/token/?scope=repository:appgate-sdp/cz-opensearch:pull&service=public.ecr.aws` See https://docs.docker.com/registry/spec/auth/scope/

but if this works, go for it.

thomascellerier: Ah I didn't do that in arc either. FYI @mandopaloooza

kajes: It works, but maybe good to add anyway

danijeel (approved):

PR #405, opened 2023-05-04
SA-21747-password-panic

kajes:
Fixes possible panic when updating a service-user with invalid input. Password is now required to be updated from either prompting or stdin.

Also makes error messages more friendly when bundling docker image for LogServer.

Fixes: SA-21747, SA-21784

### DIFF
--- a/pkg/appliance/functions.go
+++ b/pkg/appliance/functions.go
@@ -20,7 +20,6 @@ import (
 	"sync"
 
 	"github.com/appgate/sdp-api-client-go/api/v18/openapi"
-	"github.com/appgate/sdpctl/pkg/api"
 	"github.com/appgate/sdpctl/pkg/hashcode"
 	"github.com/appgate/sdpctl/pkg/network"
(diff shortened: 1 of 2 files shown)


thomascellerier (approved):

pkg/appliance/functions.go
```diff
@@ -910,12 +910,12 @@ func downloadDockerImageBundle(args imageBundleArgs) {
 		}
 		res, err := args.client.Get(fmt.Sprintf("https://public.ecr.aws/token/?scope=repository:appgate-sdp/%s:pull&service=public.ecr.aws", args.image))
 		if err != nil {
-			args.fileEntryChan <- fileEntry{err: err}
+			args.fileEntryChan <- fileEntry{err: api.HTTPErrorResponse(res, fmt.Errorf("failed to fetch public ECR token: %w", err))}
```
danijeel, line 913: `api.HTTPErrorResponse` is design with the intent to handle error response from the appgate controller, it does not really serve any purpose to use it on other domains.

PR #418, opened 2023-05-25
change LogServer image filename

kajes:
This fixes an issue where the docker image bundle would get the wrong name in the internal testing system.

Fixes: SA-21785


danijeel (approved):
a unit test would be nice here.

PR #419, opened 2023-05-29
API upgrade to v19

kajes:
Upgrade the API peer version to v19. This will deprecate support for the obsolete peer interface on appliances.


thomascellerier (approved):
In the release notes we should document the supported API versions.

kajes:
> In the release notes we should document the supported API versions.

We should probably also consider printing a warning if the appliance version is lower than supported when a command is executed.

PR #420, opened 2023-06-02
Add skip-bundle flag for skipping bundling of docker images

kajes:
Using the flag will prevent sdpctl from downloading and bundling the LogServer images when preparing. This will put the load of downloading and upgrading the LogServer image on the appliance itself.

This would be preferrable if, for example, the client that is running sdpctl, has a slow connection.

The registry which the LogServer images are pulled from needs to be accessible by the appliance in this case for upgrading of LogServer to work.

### DIFF
--- a/cmd/appliance/upgrade/prepare.go
+++ b/cmd/appliance/upgrade/prepare.go
@@ -111,7 +111,7 @@ func NewPrepareUpgradeCmd(f *factory.Factory) *cobra.Command {
 				return err
 			}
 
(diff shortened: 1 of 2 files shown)


mandopaloooza (approved):

thomascellerier:
Just a nitpick. Maybe the flag name should mention that this is about containers? E.g. `--skip-container-bundle`.

I also think we need to review the way we match the appliance image name between sdpctl and appliance upgrades. We probably want to change that to make it more resilient. We can discuss separately.

PR #434, opened 2023-06-23
Improve UX for commands annotated with updateAPIConfig

mandopaloooza:
When a command `upgrade prepare` and `upgrade complete` (commands with updateAPIConfig annotation) is run before the sdpctl configuration is initialized, it prints the following warning: 

```
It is strongly adviced that you upgrade your appliance to a supported version before executing any sdpctl command. Executing sdpctl commands on an unsupported API version can have serious unforseen consequenses.
Minimum supported API version: 17
Currently using API version: 0
``` 

sdpctl reads the key `api_version` in the config to extract the current API version. This configuration key is only initialized when the user authenticates sdpctl. I changed the orders of the warning printer after there is an attempt to signin.


thomascellerier:
I'm still not sure this warning makes sense as we always hit it when running in non interactive mode with env variables.
If the current API version is 0 we can auto discover the API version like we did in the old upgrade script, without printing a warning.

kajes (approved):

kajes:
I'm fine merging this and we should probably do a release with this fix fairly quickly.

PR #444, opened 2023-07-17
Argument order

dlnilsson:
in cases where all credentials are resolved without config file (environment variables)
we need to run these functions in the right order, first resolve the root.PersistentPreRunE (to resolve the bearer token etcetra)
after we have resolved those values, we can continue with the sub-command and correctly assert opts.Appliance() without having to worry about empty Authorization header value, which is the current case in the main branch.

This PR fixes cases where we have disabled keyring and do not persist the bearer token value, as example:

```bash
SDPCTL_URL=https://controller.appgate.devops:8443/ SDPCTL_API_VERSION=19 SDPCTL_NO_KEYRING=true SDPCTL_USERNAME=usernmaeSDPCTL_PASSWORD=the_password sdpctl appliance export-seed
```


danijeel:
Tests fails because of https://github.com/appgate/sdpctl/pull/442

kajes (approved):

PR #446, opened 2023-07-17
SA-22196 Store certificate raw in base64 format

dlnilsson:
See commit messages and youtrack issue for more info.

### DIFF
--- a/cmd/configure/configure.go
+++ b/cmd/configure/configure.go
@@ -1,16 +1,10 @@
 package configure
 
 import (
-	"crypto/sha1"
-	"crypto/sha256"
-	"crypto/x509"
 	"encoding/base64"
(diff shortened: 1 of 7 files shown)


danijeel:
Tests fails because of https://github.com/appgate/sdpctl/pull/442

gokhan-berberoglu:
very nice 🤩

kajes:
I added migration code. The `pem_filepath` is no longer set when configuring so that deprecation will be easier. Only the `pem_base64` key is set when configuring.

Migration happens on signin when the `pem_filepath` key is set but not the `pem_base64`. In which case the `pem_base64` key will be set and the value for `pem_filepath` will be set to empty.

thomascellerier (approved):

PR #460, opened 2023-08-28
Add actual-hostname flag to force-disable-controller command

kajes:
This implements the `actual-hostname` flag to the `force-disable-controller` command.

Also included in this PR is that the log will include the current status even when reaching the desired state when upgrading.


thomascellerier (approved):
Very nice :+1:

PR #463, opened 2023-08-29
Zero exit when there is nothing to prepare

kajes:
This makes the prepare command exit with code 0 if there's nothing to prepare due to appliances already prepared or already at the same version.

Error messages will still be printed to the output, but the zero exit code will not make it fail in an automated environment.

### DIFF
--- a/pkg/cmdutil/execute.go
+++ b/pkg/cmdutil/execute.go
@@ -78,7 +78,7 @@ func ExecuteCommand(cmd *cobra.Command) ExitCode {
 
 		// if error is ErrNothingToUpgrade, exit with zero code
 		if errors.Is(err, ErrNothingToPrepare) {
-			fmt.Println(err.Error())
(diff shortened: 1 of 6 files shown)


pkg/cmdutil/execute.go
```diff
@@ -76,6 +76,12 @@ func ExecuteCommand(cmd *cobra.Command) ExitCode {
 			result = multierror.Append(result, ErrCommandTimeout)
 		}
 
+		// if error is ErrNothingToUpgrade, exit with zero code
+		if errors.Is(err, ErrNothingToPrepare) {
+			fmt.Println(err.Error())
```
userappgate, line 81: doesn't this print to stdout? Shouldn't it print to stderr?

kajes: Good point! I'll fix that.

thomascellerier (approved):

PR #470, opened 2023-09-13
Improve backup downloads

kajes:
This changes how the backups are downloaded from the appliance. Backup file data is now streamed directly to the download destination and does not rely on all backup file data being downloaded before writing to the local backup file.

This makes downloading backup files more reliable and memory efficient by not relying on all data being downloaded to a temporary file before saving to the final destination.


pkg/appliance/backup/backup.go
```diff
@@ -44,16 +49,41 @@ func (b *Backup) Initiate(ctx context.Context, applianceID string, logs, audit b
...
+	res, err := client.Do(req)
+	if err != nil {
+		return nil, api.HTTPErrorResponse(res, err)
+	}
+	defer res.Body.Close()
+	if res.StatusCode > 299 {
+		return nil, api.HTTPErrorResponse(res, errors.New("unexpected response code"))
+	}
+
+	_, err = out.ReadFrom(res.Body)
```
thomascellerier, line 82: This is the part doing the streaming download right?

thomascellerier (approved):

pkg/appliance/backup/backup.go
kajes: Yep, it's reading straight from the response body, but the change on line 65 is important too for keeping the connection alive on the client until EOF.

PR #471, opened 2023-09-13
Files upload command

kajes:
This introduces a new command for uploading files to the primary controller file repository.

This simplifies the process of upgrading a Collective in a locked down environment, where sdpctl may not have access to anything else than the primary controller by making it possible to upload upgrade files separately from the upgrade command.


pkg/files/files.go
```diff
@@ -0,0 +1,92 @@
...
+	Progress *tui.Progress
+}
+
+func (f *FilesAPI) Upload(ctx context.Context, file *os.File) error {
+	fileInfo, err := file.Stat()
+	if err != nil {
+		return err
+	}
+	size := fileInfo.Size()
+	name := fileInfo.Name()
```
thomascellerier, line 27: I think it would be good to allow, optional, overriding the file name used when uploading.

So you can potentially have one file name locally and it ends up being uploaded with a different name on the controller file repository.

I have a feeling that could be useful in some support cases.

thomascellerier (approved):
Looks good but I think it would be good to also allow overriding the file name on upload.

pkg/files/files.go
kajes: Ok, it's added now. The UX is not obvious, because you can upload multiple files at once, but it works pretty well and is documented in the help.

thomascellerier (approved):
:+1:

PR #473, opened 2023-09-18
Avoid too many requests error when pulling LogServer images

kajes:
Running the `appliance upgrade prepare` command would occassionally yield in a "Too many requests" error when sdpctl is trying to pull the LogServer images from the public ECR registry.

With this PR, all requests being made to the public ECR registry will be retried if the registry responds with an error.

### DIFF
--- a/pkg/appliance/functions.go
+++ b/pkg/appliance/functions.go
@@ -915,22 +915,16 @@ func downloadDockerImageBundle(args imageBundleArgs) {
 	}
 
 	requestRetry := func(client *http.Client, req *http.Request) (*http.Response, error) {
-		var err error
-		var res *http.Response
-		err = backoff.Retry(func() error {
-			res, err = client.Do(req)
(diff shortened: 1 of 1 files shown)


pkg/appliance/functions.go
```diff
@@ -901,6 +914,25 @@ func downloadDockerImageBundle(args imageBundleArgs) {
...
+			if res.StatusCode != http.StatusOK {
+				return fmt.Errorf("Recieved %s status", res.Status)
+			}
+			return nil
+		}, backoff.NewExponentialBackOff())
+		if err != nil {
+			return nil, err
+		}
+		return res, nil
+	}
```
mandopaloooza, line 934: I'm not quite sure who this library works but shouldn't we set a max retry?

kajes: Library doesn't really have a max retry, but a default timeout (which we can control, of course) of 15 minutes, which I think is fine.

PR #474, opened 2023-09-27
Minor bugfixes and improvements

kajes:
This is collection of fixes and improvements:
- tracker would, in some cases, give wrong status message during prepare
- LogServer bundle is streamed directly to file when downloading, preventing the need to use a temporary file
- Version check disabled now logs as info instead of an error
- return error when arguments are used in prepare and complete command instead of just ignoring them


thomascellerier:
> return error when arguments are used in prepare and complete command instead of just ignoring them

I think that should be extended to *all* commands in sdpctl.
Extra arguments should never be ignored silently.

thomascellerier (approved):

PR #476, opened 2023-10-02
Place log file in correct directory

kajes:
Since "profiles" were implemented, the sdpctl log files have been placed in the wrong directory, alongside the configuration. This PR fixes this and also migrates log files from old versions and places the logs at the correct destination.

Fixes SA-22429

### DIFF
--- a/pkg/profiles/profile.go
+++ b/pkg/profiles/profile.go
@@ -9,9 +9,10 @@ import (
 
 	"github.com/appgate/sdpctl/pkg/filesystem"
 	"github.com/appgate/sdpctl/pkg/util"
+	"github.com/hashicorp/go-multierror"
 )
 
-func GetStorageDirectory() string {
+func GetConfigDirectory() string {
 	p, err := Read()
 	if err != nil {
 		return filesystem.ConfigDir()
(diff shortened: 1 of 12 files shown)


pkg/profiles/profile.go
```diff
@@ -22,12 +23,27 @@ func GetStorageDirectory() string {
...
+}
+
 func FilePath() string {
 	return filepath.Join(filesystem.ConfigDir(), "profiles.json")
 }
 
-func Directories() string {
-	return filepath.Join(filesystem.ConfigDir(), "profiles")
+func Directories() (string, string) {
+	return filepath.Join(filesystem.ConfigDir(), "profiles"), filepath.Join(filesystem.DataDir(), "logs")
```
thomascellerier, line 46: @userappgate Where should the logs go according to XDG?
I have see the following mentioned as the "right" place: `$XDG_DATA_HOME`, `$XDG_STATE_HOME`, `$XDG_CACHE_HOME`.

https://superuser.com/questions/1293842/where-should-userspecific-application-log-files-be-stored-in-gnu-linux

userappgate (approved):

pkg/profiles/profile.go
kajes: Apparently there is no "right" place, there doesn't seem to be any agreement on this.

PR #477, opened 2023-10-02
Trigger ZTP update on upgrade complete

kajes:
Trigger a ZTP version update once an upgrade is completed. This will help ZTP get the correct state of appliances after an upgrade has completed.

Fixes SA-22348

### DIFF
--- a/cmd/appliance/upgrade/complete.go
+++ b/cmd/appliance/upgrade/complete.go
@@ -814,12 +814,12 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
 	// From v18 and up
 	// This step is not fatal, so we only log errors here
 	if opts.Config.Version >= 18 {
-		ZTPStatus, err := a.ZTPStatus(ctx)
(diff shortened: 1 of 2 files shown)


cmd/appliance/upgrade/complete.go
```diff
@@ -810,6 +810,23 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
...
 
+	// Trigger ZTP version update if needed
+	// From v18 and up
+	// This step is not fatal, so we only log errors here
+	if opts.Config.Version >= 18 {
+		ZTPStatus, err := a.ZTPStatus(ctx)
+		if err != nil {
+			log.WithError(err).Warn("failed to get ZTP registered status")
+		}
+		if ZTPRegistered, ok := ZTPStatus.GetRegisteredOk(); err == nil && ok {
```
thomascellerier, line 821: Shouldn't you skip this part if ZTPStatus is nil?
Just a style thing but is it normal to name variables with leading uppercase like this? Looked like a type to me at first.

PR #478, opened 2023-10-04
fix function names for windows

kajes:
Builds are failing for windows because one of the function definititions changed when fixing the log path.


userappgate (approved):
Just be aware that I didn't actually test this :D

PR #481, opened 2023-10-09
Keep LogServer bundle after prepare

kajes:
Previously, the LogServer bundle was removed from the controller file repository once the `upgrade prepare` command was completed successfully.

We need to keep this file until after the upgrade is complete, so this PR moves the removing of the logserver bundle until `upgrade complete` has completed successfully.


kajes:
I also disabled an unpredictable test of the profiles list command

thomascellerier (approved):

PR #513, opened 2024-01-09
Switch partition command

kajes:
This PR introduces a new `switch-partition` command, which is independent from the upgrade process. It can be called by running `sdpctl appliance switch-partition [<appliance-id>]`, with the appliance-id as an optional argument. If no argument is provided, sdpctl will prompt for an appliance to make the partition switch on. Only one appliance can have it's partition switch at a time.

Fixes SA-20513


cmd/appliance/switch_partition.go
```diff
@@ -0,0 +1,225 @@
...
+
+	if p != nil {
+		p.Wait()
+	}
+
+	// verify partition switch
+	stats, _, err := api.Stats(ctx, nil, nil, false)
+	if err != nil {
+		return fmt.Errorf("partition switch failed: %w", err)
+	}
```
mandopaloooza, line 191: This error message might be a little misleading. It's not the switch-partition failing but the status call failing

cmd/appliance/switch_partition.go
```diff
@@ -0,0 +1,225 @@
...
+	var newVolume float32
+	for _, a := range stats.GetData() {
+		if a.GetId() == appliance.GetId() {
+			newVolume = a.GetVolumeNumber()
+		}
+	}
+
+	if newVolume == volume {
+		return fmt.Errorf("partition switch failed: volume number is the same as before executing the command")
```
mandopaloooza, line 200: Maybe log the volume number?

kajes: Added a log above

mandopaloooza (approved):

PR #514, opened 2024-01-09
Fix configure not checking protocol

kajes:
This PR fixes an issue where the configure command would error if the url provided in the argument does not contain the protocol as a prefix.

### DIFF
--- a/cmd/configure/configure.go
+++ b/cmd/configure/configure.go
@@ -54,7 +54,7 @@ func NewCmdConfigure(f *factory.Factory) *cobra.Command {
 			switch len(args) {
 			case 0:
 				if noInteractive || !opts.CanPrompt {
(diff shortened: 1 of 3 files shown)


cmd/configure/configure.go
```diff
@@ -140,6 +141,10 @@ func argValidation(cmd *cobra.Command, args []string) error {
 		if regex.MatchString(arg) {
 			return fmt.Errorf("'%s' is not a valid argument. Did you mean 'signin'?", arg)
 		}
+		// If arg is missing protocol prefix, temporarily add one to validate the url
+		if !strings.HasPrefix(arg, "http") {
```
thomascellerier, line 145: Maybe better to check for http:// or https://? I don't know if there is any website out there that starts with http but technically you could have a website like `httpmycontroller.example.com`.

thomascellerier (approved):

PR #520, opened 2024-01-18
Resumable backup downloads

kajes:
From appliance version 6.3, the API for downloading backups will support making ranged requests, meaning we can resume downloading a backup file if it, for some reason, fails in the middle.

This will add robustness for downloading big backup files and save time and frustration in case of bad connections either between sdpctl and the controller or between appliances themselves.

Fixes: SA-23049


kajes:
So is this fine? @thomascellerier @mandopaloooza

pkg/appliance/backup/backup.go
```diff
@@ -86,6 +88,70 @@ func (b *Backup) Download(ctx context.Context, applianceID, backupID, destinatio
...
+	if err := retryDownload(ctx, client, out, url, size); err != nil {
+		return nil, err
+	}
+
+	return out, nil
+}
+
+func retryDownload(ctx context.Context, client *http.Client, f *os.File, url string, size int64) error {
+	start := int64(0)
+	buf := make([]byte, size)
```
thomascellerier, line 127: Does that mean you are allocating a buffer the size of the whole file?
Wouldn't it be better to read in chunks and write directly to the file without accumulating everything in memory.
While still keeping track of `start` for purposes of the range header.

thomascellerier (commented):
The range retry logic looks good but I don't think we should be allocating a buffer for the whole file in memory.

pkg/appliance/backup/backup.go
```diff
@@ -86,6 +88,70 @@ func (b *Backup) Download(ctx context.Context, applianceID, backupID, destinatio
...
+	return backoff.Retry(func() error {
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, http.NoBody)
+		if err != nil {
+			return backoff.Permanent(err)
+		}
+		req.Header.Set("Range", fmt.Sprintf("bytes=%d-", start))
+
+		res, err := client.Do(req)
```
mandopaloooza, line 135: Personally, I would like to see DEBUG logging here to show what range we are currently fetching, in case we have to troubleshoot issues in the future.

kajes: I can add a log here, but running with the `--debug` flag will log every request/response anyway. It's part of the client we use.

pkg/appliance/backup/backup.go
kajes: I agree I'll see what I can do here.

pkg/appliance/backup/backup.go
mandopaloooza: `--debug` shows the request/response but it doesn't display the range headers

```
2024/01/25 07:33:41
GET /admin/appliances/ec806b3b-0385-4402-afee-12928087806b/backup/f827bfce-bb7d-11ee-82db-005056b51bc5/status HTTP/1.1
Host: envy-10-98-136-2.devops:8443
User-Agent: sdpctl/0.0.0-dev
Accept: application/vnd.appgate.peer-v20+json
Authorization: Bearer REDACTED
Accept-Encoding: gzip
```

[sdpctl-appliance-backup.log](https://github.com/appgate/sdpctl/files/14051681/sdpctl-appliance-backup.log)

pkg/appliance/backup/backup.go
```diff
@@ -86,6 +90,73 @@ func (b *Backup) Download(ctx context.Context, applianceID, backupID, destinatio
...
+			return err
+		}
+		log.WithField("response", res).Debug("download request and response")
+		defer res.Body.Close()
+		if res.StatusCode >= 400 {
+			return fmt.Errorf("response does not indicate success: %v", res.Status)
+		}
+		buf := &bytes.Buffer{}
+		n, nerr := io.Copy(buf, res.Body)
+		_, err = f.WriteAt(buf.Bytes(), start)
```
thomascellerier, line 148: Is it possible to copy directly from the response to the file and skip the buffer altogether.
I assume the write call will return the number of bytes written, which is enough to advance start.

pkg/appliance/backup/backup.go
```diff
@@ -86,6 +90,73 @@ func (b *Backup) Download(ctx context.Context, applianceID, backupID, destinatio
...
+			return backoff.Permanent(err)
+		}
+		req.Header.Set("Range", fmt.Sprintf("bytes=%d-", start))
+
+		log := logrus.WithField("request", req)
+		res, err := client.Do(req)
+		if err != nil {
+			return err
+		}
+		log.WithField("response", res).Debug("download request and response")
```
thomascellerier, line 141: Will that log the range header?

mandopaloooza: Verified log.
Im being really picky, but the colors don't work on DEBUG log for some reason. 

Screenshot 2024-01-25 at 8 45 21 AM

kajes: yep

pkg/appliance/backup/backup.go
kajes: The write call has a byte slice as a parameter, so it still needs to be copied from the body. And yes, it returns the number of bytes written, but we keep track of that in the copy call already and determine where it should start writing with that.

thomascellerier: Hmm ok i get the impression that will copy the entire body at once instead of doing it in chunks.
So a 10GB backup file would be fully be copied to the buffer in memory assuming no disconnects.

thomascellerier: Ping about my last comment here.

kajes: It will keep a buffer in memory until an error occurs. If that happens, it will still write to the file and retry with a new buffer. Supposedly the old buffer will be garbage collected by then, freeing up the memory.

This is necessary because the API still does not support doing multiple concurrent ranged requests, so this will not download the backup in chunks, but it will try to resume the download in case of a failure. Unfortunately this makes it necessary to keep a big buffer in memory. At least until we're able to make concurrent requests. Natan is aware of this, so a fix for that will come in an API update later.

thomascellerier: I am not talking about doing chunks at the API level but rather reading in chunks from the socket level.
E.g. reading 1MB at a time (or the appropriate chunk size) from the socket and writing one chunk at a time to the file.
This way you only need a 1MB buffer regardless of the size of the backup, and regardless of how the API works.

If I understand how io.Copy works this would mean calling it directly from res.Body to f, something like `io.Copy(res.Body, f)` instead of using an intermediate buffer. And use the return value to increment the offset, see https://pkg.go.dev/io#Copy

See also https://stackoverflow.com/questions/11692860/how-can-i-efficiently-download-a-large-file-using-go

kosmas-valianos: I think that is the defacto way to avoid the excessive allocation of memory. The simple way is to use io.ReadAll() but the main drawback is that it reads the full input into memory, which may cause memory exhaustion and crashes if the input is too big.

https://betterprogramming.pub/the-hidden-risks-of-a-popular-way-to-read-http-requests-in-go-208a47481f8b
>The io.Copy function can be used as an alternative to io.ReadAll to copy data from an input source to an output source more efficiently by using less memory. This is a case to consider when we need to read files from an HTTP request. Because io.Copy reads data in small chunks, it performs better and lowers the possibility of running out of memory. It uses a 32-Kilobyte buffer to store data and flushes the contents of the buffer to disk.

kajes: Yes, but `Copy` will always start to write from byte 0, meaning it will overwrite the bytes already written to file if it is called more than once and the backup will be corrupted for sure.

The stackoverflow thread concerns downloading large files efficiently, but not handling if the download fails midway. It just assumes it will, but midway download failures is the whole reason we're fixing this issue now. That's the reason why we need to use `WriteAt` in the first place, and that (and all other offset writers in go) needs a buffer.

kosmas-valianos: What can and should be be fixed is to stream the write though! Check the example in the link above

kosmas-valianos (approved):
Let's Go

pkg/appliance/backup/backup.go
kajes: > Verified log. Im being really picky, but the colors don't work on DEBUG log for some reason.

It shows correct for me though. Maybe your shell is just rendering `DEBUG` in the same color as the default?

thomascellerier (approved):

PR #527, opened 2024-02-05
Always have the same name for logserver bundle on appliance

kajes:
This PR will change the behaviour when uploading a local logserver bundle. Instead of uploading the bundle with the same name as the local file, the bundle will have a consistent `logserver-X.X.zip` name once uploaded to the appliance.

This will prevent the upgrade hook misidentifying the bundle when preparing an upgrade.

### DIFF
--- a/cmd/appliance/upgrade/prepare.go
+++ b/cmd/appliance/upgrade/prepare.go
@@ -530,6 +530,10 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions) 
 				"Content-Type":        writer.FormDataContentType(),
 				"Content-Disposition": fmt.Sprintf("attachment; filename=%q", zipInfo.Name()),
 			}
+			log.WithFields(log.Fields{
+				"local-name":  localLogServerBundleName,
+				"remote-name": logServerZipName,
+			}).Info("uploading local logserver bundle")
 			if !opts.ciMode {
 				err = uploadWithProgress(ctx, pr, "uploading "+localLogServerBundleName, zipInfo.Size(), headers)
 			} else {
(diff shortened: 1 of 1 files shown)


cmd/appliance/upgrade/prepare.go
```diff
@@ -530,7 +531,7 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions)
 				"Content-Disposition": fmt.Sprintf("attachment; filename=%q", zipInfo.Name()),
 			}
 			if !opts.ciMode {
-				err = uploadWithProgress(ctx, pr, "uploading "+logServerZipName, zipInfo.Size(), headers)
+				err = uploadWithProgress(ctx, pr, "uploading "+localLogServerBundleName, zipInfo.Size(), headers)
```
thomascellerier, line 534: Small nitpick: It would be nice to log "uploading Appgate-SDP-6.2-Logserver.zip as logserver-6.2.zip" to make clear what's happening, at least if the name is different locally

thomascellerier (approved):
Tested an upgrade from 6.1 to 6.2 using the official container bundle without renaming the file.

cmd/appliance/upgrade/prepare.go
```diff
@@ -529,8 +530,12 @@ func prepareRun(cmd *cobra.Command, args []string, opts *prepareUpgradeOptions)
 				"Content-Type":        writer.FormDataContentType(),
 				"Content-Disposition": fmt.Sprintf("attachment; filename=%q", zipInfo.Name()),
 			}
+			log.WithFields(log.Fields{
+				"local-name":  localLogServerBundleName,
+				"remote-name": logServerZipName,
+			}).Info("uploading local logserver bundle")
```
thomascellerier, line 536: nice :+1:

thomascellerier (approved):

PR #528, opened 2024-02-06
internal: add docker registry to build

kajes:
Adds option to add a docker registry as an environment variable, `DOCKER_REGISTRY_URL`, when building with `make build`


thomascellerier (approved):
Neat, one less thing to worry about.

PR #535, opened 2024-02-19
fix download so it's clear it's creating a directory

kajes:
When using the `sdpctl appliance functions download` command, there's a flag to specify a `--save-path` where the bundle should be saved.

It's not clear enough that this refers to a directory, which will be created if it does not already exist. These changes will make it more obvious that it is a directory by prompting the user to create it if it doesn't already exists.

The flag `--save-path` is deprecated with these changes and a new `--destination` is set to replace it. The old `--save-path` flag will still work, but a warning message will be printed if it is used.

In case the command is run with the `--no-interactive` flag, the default is still that the directory will be created, but messages will be printed to make it more clear as well.

This will also print the location on filesystem where the downloadable bundles are saved as the end message in the spinner once completed.


thomascellerier (approved):
Thanks.

Maybe the flag should be called `--download-dir` to reflect the behavior.
I agree with @perleino that "path" makes it sounds like the full path.

kajes (commented):
> Thanks.

Maybe the flag should be called `--download-dir` to reflect the behavior.
I agree with @perleino that "path" makes it sounds like the full path. 

Yes, it's named a bit unfortunate. We can change the flag, but maybe keep the old one as an alias for a while just in case. And `--destination` is a bit better because we use it in some other place already.

cmd/appliance/functions/download.go
```diff
@@ -139,6 +141,26 @@ func NewApplianceFunctionsDownloadCmd(f *factory.Factory) *cobra.Command {
...
+				if f.CanPrompt() {
+					if err := prompt.SurveyAskOne(&survey.Confirm{
+						Message: fmt.Sprintf("Directory '%s' does not exist. Do you want to create it now?", opts.destination),
+					}, &createDir); err != nil {
+						return err
+					}
+				} else {
+					fmt.Fprintf(opts.out, "Directory '%s' does not exist\n", opts.destination)
+				}
+				if createDir {
```
perleino, line 156: createDir always true?

kajes: no, not if the user says 'no' in the prompt. Setting it to true just makes it the default.

perleino: ah.. its used as a reference there. i did not see that

PR #552, opened 2024-04-25
Only show disk space warning when files are actually being uploaded

kajes:
This PR contains changes for handling the preparation of upgrades.

Currently, there's is a disk space warning that is triggered if sdpctl detects that an appliance has low disk space. This is to make sure that the appliances has enough disk space to handle the uploads.

The disk space warning, however, does not take into consideration that uploads may be skipped if the file already exists on the appliance, making it obsolete if no files are actually being uploaded.

This PR fixes this by moving the disk space warning check until after we know that files need to be uploaded.

### DIFF
--- a/cmd/appliance/files/upload.go
+++ b/cmd/appliance/files/upload.go
@@ -81,7 +81,7 @@ func NewFilesUploadCmd(f *factory.Factory) *cobra.Command {
 			}
 
 			ctx := context.Background()
-			filesAPI := files.FilesAPI{API: api}
+			filesAPI := files.FileManager{API: api}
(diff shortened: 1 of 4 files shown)


thomascellerier (approved):
Note that we still might need disk space on Controllers for several reasons:
- We usually take a backup of the primary controller, which requires some temporary storage.
- We take a copy of the database to be able to rollback a failed upgrade. Note that usually this only overwrites an old copy of the database from a previous upgrade to I don't think its much of an issue in practice.

I still think this change makes sense in general though as we have too many false positives before it.

PR #558, opened 2024-05-15
additional check for changed hostname resolution

kajes:
Adds an additional check just when upgrading the primary controller to detect change in hostname resolution during the upgrade.

If a change is detected, the upgrade will fail and detailed errors will be printed to the log.

### DIFF
--- a/cmd/appliance/upgrade/complete.go
+++ b/cmd/appliance/upgrade/complete.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 	"text/template"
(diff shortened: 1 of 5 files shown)


cmd/appliance/upgrade/complete.go
```diff
@@ -547,14 +549,34 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
...
 			}
 
-			logEntry := log.WithField("appliance", controller.GetName())
+			logEntry := log.WithFields(log.Fields{
+				"appliance": controller.GetName(),
+				"url":       cfg.URL,
+			})
+			ips, err := network.ResolveHostnameIPs(cfg.URL)
+			if err != nil {
+				logEntry.WithError(err).Error("failed to lookup hostname ips")
```
thomascellerier, line 558: Maybe this should just be a warning?
Since its an extra check we could continue the upgrade anyways.

thomascellerier (approved):

cmd/appliance/upgrade/complete.go
kajes: we do, it's just logged as an error in the logs, but execution continues after that anyway.

mandopaloooza (approved):

PR #567, opened 2024-06-24
Feature: show name in resolve-name-status table

kajes:
This PR is an update for the resolve-name-status command. The update includes showing the resolved name in the table presented instead of only in the json output.

Partial resolutions are excluded from the table by default to not clutter the output, but can be shown using the `--partial-resolutions`

### DIFF
--- a/cmd/appliance/resolve_name_status.go
+++ b/cmd/appliance/resolve_name_status.go
@@ -18,13 +18,14 @@ import (
 )
 
 type resolveNameStatusOpts struct {
-	Config      *configuration.Config
-	Out         io.Writer
-	Client      func(c *configuration.Config) (*openapi.APIClient, error)
-	Appliance   func(c *configuration.Config) (*appliancepkg.Appliance, error)
-	debug       bool
-	json        bool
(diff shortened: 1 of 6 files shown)


thomascellerier (approved):
Note that this API is changing quite a lot in 6.4 so sdpctl will need to be updated to handle the differences in API version 21.

The new format is as follows (in python dataclasses terms):
```Python
@dataclass
class NameResolution:
    ips: list[str]
    errors: list[str]


@dataclass
class NamesStatusResponse:
    resolutions: dict[str, NameResolution]
```

kajes:
> Note that this API is changing quite a lot in 6.4 so sdpctl will need to be updated to handle the differences in API version 21.
> 
> The new format is as follows (in python dataclasses terms):
> 
> ```python
> @dataclass
> class NameResolution:
>     ips: list[str]
>     errors: list[str]
> 
> 
> @dataclass
> class NamesStatusResponse:
>     resolutions: dict[str, NameResolution]
> ```

Alright, I'll make sure it gets an update by then :+1:

PR #585, opened 2024-09-04
fix high memory usage when bundling logserver image

kajes:
When upgrading a LogServer, the prepare includes downloading and bundling up the LogServer image for upload to the LogServer.

The bundling would store temporary files in memory until all files had been written to the bundle. This results in very high memory usage during the process.

This PR fixes the high memory usage issue by storing the files that should be bundled in temporary files on disk until they are written to the bundle. This decreases the memory usage significantly.

### DIFF
--- a/pkg/appliance/functions.go
+++ b/pkg/appliance/functions.go
@@ -832,7 +832,7 @@ type ImageJSON struct {
 }
 
 type fileEntry struct {
-	r    io.Reader
+	r    *os.File
 	path string
 	err  error
 }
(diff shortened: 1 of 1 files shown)


pkg/appliance/functions.go
```diff
@@ -1020,10 +1056,12 @@ func downloadDockerImageBundle(args imageBundleArgs) {
 		args.fileEntryChan <- fileEntry{err: err}
 		return
 	}
+	tmpManifest.Close()
```
mandopaloooza, line 1059: Do we need a `defer` so the file gets closed even if there is an exception before it gets closed?

kajes: Actually we need to close it before it's sent on the channel, because it's going to be opened again once it's sent. And it would still not run if there's a panic of some kind even if it's deferred, so it's not guaranteed to be closed anyway.

mandopaloooza (approved):

PR #586, opened 2024-09-06
Fix: panic on `upgrade complete`

kajes:
If there are multiple controllers in the collective, but only one is left to upgrade, the `upgrade complete` command would panic with an "index out of range" issue.

This fixes the issue.


mandopaloooza (approved):

thomascellerier (approved):
Would it be possible to write a test for this?

kajes (commented):
> Would it be possible to write a test for this? 

Added a test for this now

PR #589, opened 2024-09-12
Feature: new sites command

kajes:
This introduces a new sites command in sdpctl, used by executing `sdpctl sites`.

Initially it only supports listing sites (i.e. `sdpctl sites list`). Sites can be listed either in table format or in json format using the  `--json` flag.

A user can also query for specific sites using arguments. Arguments can be either a site name or a site id.

Examples:
```
>sdpctl sites list
>sdpctl sites list --json
>sdpctl sites list <SITE_NAME|SITE_ID>
```


thomascellerier (approved):
Looks good, thanks!

PR #596, opened 2024-10-10
Add version constraint for breaking upgrade path

kajes:
Upgrading from appliance version 6.3.5+ to 6.4.0 is known to be breaking.

This PR implements a check to restrict this upgrade. A new function is created that can be used to catch other breaking upgrade paths in the future.


thomascellerier (approved):
Thanks!

PR #597, opened 2024-10-11
fix: html escape characters in backup password

kajes:
When enabling the backup API, the user is prompted to enter a password for encrypting the backup in a prompt.

If the user enters a password containing the characters '<', '>' and '&', those characters would be escaped in the json encoding before sending the request to the appliance.

This would result in the password not being what the user expects it to be, making the backup practically impossible for user to decrypt.

This PR overrides the default behaviour of the json encoder to not escape these characters before sending the request payload.

### DIFF
--- a/cmd/appliance/backup/api.go
+++ b/cmd/appliance/backup/api.go
@@ -1,10 +1,14 @@
 package backup
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
+	"strings"
 
 	"github.com/appgate/sdp-api-client-go/api/v21/openapi"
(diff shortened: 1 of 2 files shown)


userappgate:
This was not what the bug was about, but good to fix this as well!

thomascellerier (approved):
Looks like a valuable fix yes.

Could you add some unit tests for this before merging, feels like the kind of thing that's easy to unit test and quite dangerous if it happens in practice, as in customers can't decrypt their own backups.

userappgate:
@thomascellerier maybe a simple change to the e2e test to use one of those characters in the passphrase could be useful? If we have an e2e that uses sdpctl to set the passphrase.

thomascellerier:
Good suggestion, i think we do that for the backup/restore test and the appliance upgrade test.

PR #598, opened 2024-10-11
fix: force-disable-controller selecting extra controllers

kajes:
I some cases, the force-disable-controller command would select more controllers than was selected in the selection prompt. This would happen when two controllers would have the same hostname set, leading to the command selecting all controllers with a matching hostname for disabling.

This PR fixes the issue by matching the controllers on the ID instead of the hostname.

### DIFF
--- a/cmd/appliance/force_disable_controller.go
+++ b/cmd/appliance/force_disable_controller.go
@@ -155,6 +155,7 @@ func forceDisableControllerRunE(opts cmdOpts, args []string) error {
 	sort.SliceStable(offline, func(i, j int) bool { return offline[i].GetName() < offline[j].GetName() })
 	sort.SliceStable(statData, func(i, j int) bool { return statData[i].GetName() < statData[j].GetName() })
 
+	idList := []string{}
 	unselectedOffline := []string{}
 	if len(args) <= 0 {
 		selectable := []string{}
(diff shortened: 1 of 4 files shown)


mandopaloooza (commented):
Hmm Im not sure if having the same hostname was the issue. I don't think we even allow that in the SDP configuration. Below is the snippet from the support case. 
```
This will force disable the selected controllers and announce it to the remaining controllers. The following Controllers are going to be disabled:

Name                                    Hostname                       Status     Version
----                                    --------                       ------     -------
AppGate SecondaryController - AWS       controllerdcaws.example.com    error      6.3.2-37826-release
AppGate SecondaryController - AWS DR    controllerdraws.example.com    warning    6.3.3-38086-release

? Do you want to continue? Yes

 - disabling controllers: disabling
```

Note that the hostname looks very similar, but they are not the same.

thomascellerier:
Yeah we might need some more fix here.

cmd/appliance/force_disable_controller.go
```diff
@@ -220,27 +220,26 @@ func forceDisableControllerRunE(opts cmdOpts, args []string) error {
...
 			log.WithField("arg", arg).Info("No Controller found with provided hostname")
 		}
 		// automatically add offline controllers to disable list, but only if the user has actively selected a valid controller first
-		if len(hostnameArgs) > 0 {
+		if len(idList) > 0 {
 			for _, ctrl := range offline {
-				hostnameArgs = util.AppendIfMissing(hostnameArgs, ctrl.GetHostname())
+				idList = util.AppendIfMissing(idList, ctrl.GetId())
 			}
 		}
```
mandopaloooza, line 240: I wonder if there is a bug when we compute the `offline` appliances that would automatically add non-healthy appliances. 

in `FilterAvailable` 
https://github.com/appgate/sdpctl/blob/4fc40b114ffa84afcf73b238ea92523acc627b3b/pkg/appliance/functions.go#L124

mandopaloooza: Hmmm maybe not. This looks correct.
https://github.com/appgate/sdpctl/blob/4fc40b114ffa84afcf73b238ea92523acc627b3b/pkg/appliance/functions.go#L795

kajes:
I'll merge this one even if it may not be the culprit here, but I think this makes it a bit more robust.

PR #601, opened 2024-10-22
Extra upgrade status checks

kajes:
This adds an extra check for upgrade status of the appliances just before performing the upgrade just in case the appliance goes offline or any other error occurs after the upgrade is initialized.

### DIFF
--- a/cmd/appliance/upgrade/complete.go
+++ b/cmd/appliance/upgrade/complete.go
@@ -417,22 +417,30 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
 		}
 		for _, appliance := range appliances {
 			i := appliance
+			logger := log.WithFields(log.Fields{
+				"appliance": i.GetName(),
+				"id":        i.GetId(),
+			})
 			var initialVolume float32
 			for _, appData := range initialStats.GetData() {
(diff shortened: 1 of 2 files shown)


cmd/appliance/upgrade/complete.go
```diff
@@ -417,36 +417,55 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
...
+			logger.WithField("volume", initialVolume).Info("registered initial volume number")
 			g.Go(func() error {
 				ctx, cancel := context.WithTimeout(ctx, opts.Timeout)
 				defer cancel()
-				logEntry := log.WithField("appliance", i.GetName())
-				logEntry.Info("checking if ready")
+				logger := log.WithFields(log.Fields{
+					"appliance": i.GetName(),
+					"id":        i.GetId(),
+				})
```
mandopaloooza, line 437: Can we reuse the same logger declared in line 420?

mandopaloooza (approved):
Just one comment about the logger.

We should also start inviting @torgermiller and @ravenc-appgate to sdpctl code reviews

cmd/appliance/upgrade/complete.go
kajes: yes we can :+1:

PR #602, opened 2024-10-22
fix: set output before logging anything

kajes:
Set the log output before anything is logged to prevent ci-mode from breaking


thomascellerier (approved):
sneaky

PR #608, opened 2024-11-06
internal: disregard pre-release strings in version

kajes:
Only build number will be considered when semvers are the same.


thomascellerier (approved):
Thanks!

PR #609, opened 2024-11-07
New command to process .journal files and create .log text files

userappgate:
This will extract all the .journal files from the .zip file and will do the grouping by syslog identifier like we currently do server-side.

`sdpctl  appliance extract-logs controller-0421a864-f171-46f2-b059-f8c35acef9ff-site1-logs.zip  --path iaffieddu/`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ dist/
 cover.out
 .idea/
 .release-env
+.kdev4/
+sdpctl.kdev4
--- a/cmd/appliance/extract_logs.go
+++ b/cmd/appliance/extract_logs.go
@@ -28,7 +28,7 @@ func NewExtractLogsCmd(f *factory.Factory) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "extract-logs",
 		Short:   docs.ApplianceExtractLogsDoc.Short,
-		Long:    docs.ApplianceExtractLogsDoc.Short,
+		Long:    docs.ApplianceExtractLogsDoc.Long,
 		Example: docs.ApplianceExtractLogsDoc.ExampleString(),
 		RunE: func(c *cobra.Command, args []string) error {
 			return logsExtractRun(c, args, &opts)
--- a/cmd/appliance/extract_logs.go
+++ b/cmd/appliance/extract_logs.go
@@ -52,7 +52,7 @@ func NewExtractLogsCmd(f *factory.Factory) *cobra.Command {
 
 func logsExtractRun(cmd *cobra.Command, args []string, opts *logextractOpts) error {
 	for i := 0; i < len(args); i++ {
-		fmt.Println("Starting processing %s", args[i])
+		log.Infof("Starting processing %s", args[i])
 		err := process_journal_file(args[i], opts.Path)
 		if err != nil {
 			return err
@@ -74,7 +74,7 @@ func process_journal_file(file string, path string) error {
 	// printing some of their contents.
 	for _, f := range r.File {
 		if strings.HasSuffix(f.Name, ".journal") {
-			fmt.Println("Extracting journal file %s", f.Name)
+			log.Infof("Extracting journal file %s", f.Name)
 
 			rc, err := f.Open()
 			if err != nil {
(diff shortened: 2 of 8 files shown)


cmd/appliance/extract_logs.go
```diff
@@ -0,0 +1,185 @@
...
+				return err
+			}
+			extractedFiles = append(extractedFiles, extracted.Name())
+
+			written, err := io.Copy(extracted, rc)
+			if written+1 == written-1 { // Completely useless statement to get go to compile
+				log.Fatal(written)
+			}
+			if err != nil {
+				log.Fatal(err)
```
kajes, line 94: This will terminate the program. I think it would be better to return the error or collect the errors that occur in the loop and return that.

cmd/appliance/extract_logs.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	opts := logextractOpts{
+		aopts,
+		"",
+	}
+	cmd := &cobra.Command{
+		Use:     "extract-logs",
+		Short:   docs.ApplianceLogsDoc.Short,
+		Long:    docs.ApplianceLogsDoc.Short,
```
kajes, line 39: Some proper documentation would be nice, unless it's supposed to replace the current `appliance logs` command.

kajes (commented):
I think this looks fine. I'm just worried that it will be confusing to have two different log commands. Is this supposed to replace the current `appliance logs` command?

.gitignore
```diff
@@ -4,3 +4,5 @@ dist/
 cover.out
 .idea/
 .release-env
+.kdev4/
+sdpctl.kdev4
```
kajes, line 8: What are these for, and should they be commited?

cmd/appliance/extract_logs.go
```diff
@@ -0,0 +1,185 @@
...
+			extracted, err := os.CreateTemp("", "")
+			if err != nil {
+				return err
+			}
+			extractedFiles = append(extractedFiles, extracted.Name())
+
+			written, err := io.Copy(extracted, rc)
+			if written+1 == written-1 { // Completely useless statement to get go to compile
+				log.Fatal(written)
+			}
```
kajes, line 92: It doesn't compile because you're declaring a variable that you are not using, so this is the correct way of doing it in go:
```suggestion
			_, err := io.Copy(extracted, rc)
```
`_` marks a return that should be ignored.

.gitignore
userappgate: They are for my IDE, Yes they should be like all the other things to ignore stuff from other IDE.

cmd/appliance/extract_logs.go
kajes: Oh, yeah, you're re-defining the `err` variable as well, which means `=` instead of `:=`, so this is the way:
```suggestion
			_, err = io.Copy(extracted, rc)
```

kajes (approved):
Looks good now, so merge when you're ready :+1:

kajes (commented):
Found a small thing that probably should be commited before merge.

cmd/appliance/extract_logs.go
```diff
@@ -0,0 +1,174 @@
...
+				textlogs[identifier] = logfile
+			}
+
+			logfile.WriteString(formatEntry(entry))
+		}
+
+		j.Close()
+		if os.Remove(journalfile) != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
userappgate: Ah right!

That doesn't work either though because it will only loop once if it works.

Go needs a walrus operator :D (or an assignment operator that also returns the assigned value like C)

cmd/appliance/extract_logs.go
```diff
@@ -0,0 +1,174 @@
...
+
+		err := j.Open(journalfile)
+		if err != nil {
+			continue
+		}
+
+		for {
+			entry, hasnext, err := j.Next()
+			if err != nil {
+				// We just move to the next log file in case of error
```
thomascellerier, line 110: Do we want a verbose mode where we log the errors here?

thomascellerier (approved):

cmd/appliance/extract_logs.go
userappgate: Normally that happens at least once because the last log will be "incomplete" as it's still open for writing when being sent.

PR #610, opened 2024-11-11
finish upgrade batch before returning error

kajes:
Currently, when an appliance upgrade fails for some reason in a batch, the command exits before the rest of the appliances in the same batch have finished the upgrade process.

This makes the user unaware on when the rest of the appliances have finished running through the upgrade.

This PR fixes the issue by finishing the whole upgrade batch before returning the errors.

### DIFF
--- a/cmd/appliance/upgrade/complete.go
+++ b/cmd/appliance/upgrade/complete.go
@@ -405,7 +405,7 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
 	}
 
 	batchUpgrade := func(ctx context.Context, appliances []openapi.Appliance, SwitchPartition bool) error {
-		g, ctx := errgroup.WithContext(ctx)
+		g := errgroup.Group{}
 		upgradeChan := make(chan openapi.Appliance, len(appliances))
 		var p *tui.Progress
 		if !opts.ciMode {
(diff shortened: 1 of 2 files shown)


cmd/configure/configure.go
```diff
@@ -142,7 +142,7 @@ func argValidation(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("'%s' is not a valid argument. Did you mean 'signin'?", arg)
 		}
 		// If arg is missing protocol prefix, temporarily add one to validate the url
-		if !strings.HasPrefix(arg, "https://") || !strings.HasPrefix(arg, "http://") {
+		if !strings.HasPrefix(arg, "https://") && !strings.HasPrefix(arg, "http://") {
```
appgate-nseigler, line 145: nice fix!

cmd/appliance/upgrade/complete.go
```diff
@@ -455,7 +455,11 @@ func upgradeCompleteRun(cmd *cobra.Command, args []string, opts *upgradeComplete
 						return nil
 					}, backoff.WithContext(backoff.NewExponentialBackOff(), ctx))
 					if err != nil {
-						return fmt.Errorf("upgrade complete failed on %s: %w", i.GetName(), err)
+						if !opts.ciMode {
+							t.Update(appliancepkg.UpgradeStatusFailed)
+							t.Fail(err.Error())
```
appgate-nseigler, line 460: I have read the code here: https://github.com/appgate/sdpctl/blob/6c4dc6ae53fcd00fd99d4da3f61f534d70a12f06/pkg/tui/tracker.go#L123-L130

I dont fully understand what this does with the select.

We take the error message and place it on the channel continuously until the program ends?

This is the only outcome of calling t.Fail right? Just reading it in this file made me think we were panicking or doing some other exit/fail now behavior but I think it is just a notification message then a sleep.

kajes: Yep, that's pretty much it. This should never be blocking anything. It's merely for presentation. We've had a lot of issues in the past where these progress bars and trackers were blocking for no apparent reason, so we've ended up with a better-safe-than-sorry approach on these to make sure they never block.

thomascellerier (approved):

PR #615, opened 2024-12-06
[SA-24689] Prevent batch size to be 0 when batch < maxUnavailable

mandopaloooza:
When `--max-unavailable` is larger than the batch size, the batch size could become 0. The batch size should always be greater than or equal to 1. 

Bonus: Fix unit test on Mac


thomascellerier (approved):
Thanks!

PR #623, opened 2025-01-29
Support zstd compression in log bundle

userappgate:


cmd/appliance/extract_logs.go
```diff
@@ -58,6 +59,11 @@ func processJournalFile(file string, path string) error {
 	}
 	defer r.Close()
 
+	r.RegisterDecompressor(93, func(in io.Reader) io.ReadCloser {
```
thomascellerier, line 62: What's 93? This should use a constant or at least a comment.

thomascellerier (commented):
Wondering about the number 93 :).

cmd/appliance/extract_logs.go
userappgate: It's the number that they assigned in the zip format to the zstd algorithm.

```
In [2]: zipfile.ZIP_ZSTANDARD
Out[2]: 93
```

I will change it

mandopaloooza (approved):

PR #631, opened 2025-02-20
Improvements on extract-logs   

userappgate:
Since the standard unzip command does not support the zstd compression algorithm that the current appliances are using, this pull request lets sdpctl with the appropriate flags to be used instead.

The main changes here are:

* All files are extracted, not only .journal files
* Files located inside a directory in the zip are placed in a directory when extracted
* A new flag is introduced to extract the .journal files without processing them. This is to be used by linux users who already have journalctl installed and want to read the logs with ease.

### DIFF
--- a/cmd/appliance/extract_logs.go
+++ b/cmd/appliance/extract_logs.go
@@ -105,7 +105,7 @@ func processJournalFile(file string, path string, unzipOnly bool) error {
 				parts = parts[:len(parts)-1]
 				prevpath := path
 
-				for i, part := range parts {
+				for _, part := range parts {
 					dirpath := filepath.Join(prevpath, part)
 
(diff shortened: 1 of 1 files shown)


torgermiller (approved):

mandopaloooza (commented):
Few comments on usability

No error message when I run with no arguments
```
❯ go run main.go appliance extract-logs

```

Extraction work but it doesn't log anything to stdout
```
❯ go run main.go appliance extract-logs ~/Downloads/controller-f99f6765-97b8-4382-92fb-5e433708f550-site1-logs.zip --path ~/Downloads/

```

userappgate:
I use the `log.Infof` thing inside to match with the rest of the codebase. It doesn't print anything by default that…

PR #633, opened 2025-02-27
[SA-22730] use on-boarded-devices API

ravenc-appgate:
Use on-boarded-devices API for token revocation instead of deprecated token-records API


mandopaloooza (commented):
Nice! Looks good to me 🚢 

I will let others review the PR as well.

torgermiller (approved):

pkg/token/api.go
```diff
@@ -19,39 +19,31 @@ type Token struct {
 	Token      string
 }
 
-func (t *Token) ListDistinguishedNames(ctx context.Context, orderBy []string, descending bool) ([]openapi.DistinguishedName, error) {
```
natan-abolafya, line 22: This file, structs etc. should probable be renamed now.

natan-abolafya (approved):
not a go developer but the changes generally look fine. IMO some token->registered-device rename should be done though.

PR #649, opened 2025-04-07
ZTP-3816 Upgrade should exclude Autoscaling Gateways

damara-appgate:
Autoscaling gateways are now excluded from appliances selected for upgrade.

### DIFF
--- a/cmd/appliance/upgrade/prepare.go
+++ b/cmd/appliance/upgrade/prepare.go
@@ -319,6 +319,7 @@ func prepareRun(cmd *cobra.Command, opts *prepareUpgradeOptions) error {
 			}
 		}
 		// Remove auto-scaled gateways from appliances to be upgraded
+		log.Info("excluding autoscaling gateways")
 		appliancesForUpgrade := make([]openapi.Appliance, 0)
 		for _, appliance := range appliances {
 			for _, gw := range gws {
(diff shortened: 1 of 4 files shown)


cmd/appliance/upgrade/prepare.go
```diff
@@ -303,24 +303,32 @@ func prepareRun(cmd *cobra.Command, opts *prepareUpgradeOptions) error {
 	}
 
 	hasLowDiskSpace := appliancepkg.HasLowDiskSpace(initialStats.GetData())
-	autoScalingWarning := false
-	constraints, _ := version.NewConstraint(">= 5.5.0")
-	if constraints.Check(opts.targetVersion) {
-		autoScalingWarning = true
-	}
```
damara-appgate, line 310: My understanding is that this check should always be performed, so I removed this constraint.

cmd/appliance/upgrade/prepare.go
```diff
@@ -303,24 +303,32 @@ func prepareRun(cmd *cobra.Command, opts *prepareUpgradeOptions) error {
...
+		log.Info("excluding autoscaling gateways")
+		appliancesForUpgrade := make([]openapi.Appliance, 0)
+		for _, appliance := range appliances {
+			for _, gw := range gws {
+				if appliance.Id != gw.Id {
+					appliancesForUpgrade = append(appliancesForUpgrade, appliance)
+				}
+			}
 		}
+		appliances = appliancesForUpgrade
```
damara-appgate, line 331: Is there a better / preferred way to subtract the elements of one list from another list in Go?

mandopaloooza: I find that working with slices in Golang feels unpleasant, especially coming from Python. I think this logic is fine as long as the logic is easy to understand, which it is. 👍

mandopaloooza (approved):
Looks good to me 🚀

PR #653, opened 2025-04-10
[SA-25515] Fix Metrics Aggregator enum and add function

ravenc-appgate:
Fix the Metrics Aggregator enum and add it as a valid appliance function.


mandopaloooza (approved):
LGTM 👍

PR #657, opened 2025-04-15
[SA-25533] Allow upgrade images to be specified by absolute path

ravenc-appgate:
The URL parsing logic was returning nil on paths starting with '/'. This caused the path to be interpreted as a URL instead of a local path.


torgermiller (approved):

mandopaloooza (approved):
Nice! Good find 🚀

PR #659, opened 2025-05-02
[SA-25601] fix version parsing for multidigit versions

ravenc-appgate:


mandopaloooza (approved):
For next time, a description of the issue and fix would be appreciated! Otherwise the changes looks good to me 👍

PR #663, opened 2025-05-06
ZTP-3932 - Exclude multiple autoscaling gateways from upgrade

damara-appgate:
A previous attempt to exclude autoscaling gateways from upgrades failed when there were multiple autoscaling gateways. Now, whether one or multiple, autoscaling gateways will be excluded from upgrades.

### DIFF
--- a/cmd/appliance/upgrade/prepare.go
+++ b/cmd/appliance/upgrade/prepare.go
@@ -322,11 +322,15 @@ func prepareRun(cmd *cobra.Command, opts *prepareUpgradeOptions) error {
 		// Remove auto-scaled gateways from appliances to be upgraded
 		log.Info("excluding autoscaling gateways")
 		appliancesForUpgrade := make([]openapi.Appliance, 0)
+		// Build a map of autoscaling gateway ids
+		autoscalingGatewayIds := make(map[*string]bool)
+		for _, gw := range gws {
+			autoscalingGatewayIds[gw.Id] = true
+		}
+		// Select appliances that do not match the id of an autoscaling gateway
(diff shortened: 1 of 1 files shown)


cmd/appliance/upgrade/prepare.go
```diff
@@ -322,11 +322,15 @@ func prepareRun(cmd *cobra.Command, opts *prepareUpgradeOptions) error {
...
+			autoscalingGatewayIds[gw.Id] = true
+		}
+		// Select appliances that do not match the id of an autoscaling gateway
 		for _, appliance := range appliances {
-			for _, gw := range gws {
-				if appliance.Id != gw.Id {
-					appliancesForUpgrade = append(appliancesForUpgrade, appliance)
-				}
+			if !autoscalingGatewayIds[appliance.Id] {
+				appliancesForUpgrade = append(appliancesForUpgrade, appliance)
```
mandopaloooza, line 333: I would have just iterated the ASG IDs and make sure the appliance isn't matching it. `map[string]bool` of IDs to Boolean seems a bit redundant to me but if it works, I'm okay with it 😄

mandopaloooza (approved):
Just a nit-picky comment about implementation. You can ignore it 😄  

🚢

mandopaloooza (approved):
Just a nit-picky comment about implementation. You can ignore it 😄  

🚢

cmd/appliance/upgrade/prepare.go
damara-appgate: That makes sense. I'll make this change. Thank you!

hydrolucid (approved):

PR #665, opened 2025-05-15
[SA-25636] translate old stats fields to new status format

ravenc-appgate:
Fallback to the old stats endpoint if we get a 404 from the new one. Translate the fields available into the new stats data structure.


pkg/appliance/api.go
```diff
@@ -141,8 +141,154 @@ func (a *Appliance) UpgradeCancel(ctx context.Context, applianceID string) error
...
+
+var warningDisplayed = false
+
 func (a *Appliance) ApplianceStatus(ctx context.Context, filter map[string]map[string]string, orderBy []string, descending bool) (*openapi.ApplianceWithStatusList, *http.Response, error) {
 	status, response, err := a.APIClient.AppliancesApi.AppliancesStatusGet(ctx).Execute()
+	if response.StatusCode == 404 {
+		if !warningDisplayed {
+			fmt.Println("WARNING: Status endpoint not found, falling back to old stats/appliances")
+			warningDisplayed = true
+		}
```
mandopaloooza, line 284: This is to only print this warning once?

hydrolucid (approved):

mandopaloooza (approved):

PR #667, opened 2025-05-28
Handle status of offline appliances gracefully

torgermiller:


mandopaloooza (approved):
LGTM. Maybe a little more description next time for those who might not have context would be nice!

PR #673, opened 2025-06-11
Add list resources command

torgermiller:


mandopaloooza:
```
❯ SDPCTL_DOCKER_REGISTRY=got02-nexus01.agi.appgate.com:5001 go run main.go sites resources 8a4add9e-0e99-4bb1-949c-c9faf9a49ad4
Querying resource names...

Name    Resolver    Type    Gateway Name
----    --------    ----    ------------
No resources found in the site
```

It spent about 30 seconds until the results came back. Does it take this long because sdpctl is waiting for the server response (i.e. it is a compute-intensive task in the Controller)?

mandopaloooza (approved):
@torgermiller Thanks for the explanation! 

I'm not necessarily looking for the fix in this PR. I think we can ship it as it is and improve later on if we receive feedback that it needs to run faster. I highly doubt anyone would complain about the speed as long as it returns the correct outputs 😄

PR #686, opened 2025-07-31
Extend appliance wait timeout

ravenc-appgate:
Extend timeout to accommodate slow connections.

### DIFF
--- a/pkg/appliance/status.go
+++ b/pkg/appliance/status.go
@@ -65,12 +65,18 @@ func (u *ApplianceStatus) WaitForApplianceState(ctx context.Context, appliance o
 	logEntry := log.WithFields(log.Fields{
 		"appliance": appliance.GetName(),
 	})
+
 	logEntry.WithField("want", want).Info("Polling for the appliance state")
+	timeoutMult := 1
 	return backoff.Retry(func() error {
-		ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+		ctx, cancel := context.WithTimeout(ctx, time.Duration(timeoutMult)*5*time.Second)
 		defer cancel()
 		stats, _, err := u.Appliance.ApplianceStatus(ctx, nil, nil, false)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/appliance/status.go
```diff
@@ -67,7 +67,7 @@ func (u *ApplianceStatus) WaitForApplianceState(ctx context.Context, appliance o
 	})
 	logEntry.WithField("want", want).Info("Polling for the appliance state")
 	return backoff.Retry(func() error {
-		ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+		ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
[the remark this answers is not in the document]
mandopaloooza: Can this somehow be made configurable? The last thing I want to happen is attempt another upgrade and this timeout doesn't work.

pkg/appliance/status.go
```diff
@@ -65,12 +65,18 @@ func (u *ApplianceStatus) WaitForApplianceState(ctx context.Context, appliance o
 	logEntry := log.WithFields(log.Fields{
 		"appliance": appliance.GetName(),
 	})
+
 	logEntry.WithField("want", want).Info("Polling for the appliance state")
+	timeoutMult := 1
 	return backoff.Retry(func() error {
-		ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+		ctx, cancel := context.WithTimeout(ctx, time.Duration(timeoutMult)*5*time.Second)
```
mandopaloooza, line 72: This will do four retries? 5, 10, 15, 20 seconds. We can never go past 20 seconds because the outer backoff has a 20 second maxInterval, right?

ravenc-appgate: No. On timeout it doubles the timeout value, retrying until the total elapsed time is 10 minutes. The backoff interval has nothing to do with the context timeout, it only specifies the amount of time between retry attempts.

mandopaloooza: Ohh I see. Thanks for clarification 👍

mandopaloooza (approved):

PR #688, opened 2025-08-05
Make message less error-like

ravenc-appgate:
Make the message less error-like when the version provided is greater than or equal to the provided version.


ravenc-appgate:
> Do you think it'd be straightforward to have this not show as an error at all but rather a "notice" or something?
> 
> Seems like they want customers to not see "error" at all when hitting this as it prompts them to open support cases.

That's what this PR is. This is what the message looks like now.
image

torgermiller:
> > Do you think it'd be straightforward to have this not show as an error at all but rather a "notice" or something?
> > Seems like they want customers to not see "error" at all when hitting this as it prompts them to open support cases.
> 
> That's what this PR is. This is what the message looks like now. image

Nice looks good!

torgermiller (approved):

PR #689, opened 2025-08-07
[ZTP-3864] Improve URL parsing

RicardoRamirezG:
Currently our URL parsing is very sensitive and doesn't handle some common cases very well.


torgermiller:
Can you add some test cases?

ravenc-appgate (approved):

PR #697, opened 2025-09-05
Filter Out Ephemeral Appliances

ravenc-appgate:
Filter out ephemeral appliances when doing a backup. These appliances store no configuration, so backing them up doesn't make sense.


mandopaloooza (commented):
What do you think of calling these appliances `Stateless` instead of `Ephemeral`?

ravenc-appgate:
> What do you think of calling these appliances `Stateless` instead of `Ephemeral`?

Good idea

PR #740, opened 2026-06-01
use batching logic for ha connectors

ravenc-appgate:


mandopaloooza (approved):
LGTM. Next time, please put the library version changes in a separate commit so it can be filtered out from actual code change.

